PR #10, opened 2026-03-11
feat: add ClawHub-compatible metadata to SKILL.md front matter

JimCollinson:
Add metadata.openclaw block declaring runtime requirements (curl for install script) so the skill can be published to ClawHub via the CLI.

### DIFF
--- a/SKILL.md
+++ b/SKILL.md
@@ -20,7 +20,7 @@ metadata:
       env: []
       bins:
         - curl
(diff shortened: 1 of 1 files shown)


dirvine:
Thanks @JimCollinson! Landed in fb1ccca — the ClawHub metadata is a nice addition for OpenClaw discoverability.

PR #87, opened 2026-05-21
docs: clarify encrypted KvStore policy status

JimCollinson:
## Summary
- clarify that AccessPolicy::Encrypted is a reserved policy shape, not currently wired encrypted KvStore sync
- remove comments implying KvStore deltas are encrypted or group-authorized today
- leave behavior and APIs unchanged

## Verification
- cargo fmt --all -- --check
- cargo test --lib --all-features kv_store (blocked locally: x0x@0.19.47 requires rustc 1.95.0; active toolchain is rustc 1.94.0)

<h3>Greptile Summary</h3>

This PR corrects misleading documentation in the `KvStore` module by accurately reflecting that `AccessPolicy::Encrypted` is a reserved policy shape — not a currently functional encryption mechanism. No behavior, APIs, or logic are changed.

- `src/kv/store.rs`: Updates the module-level doc comment, the `AccessPolicy::Encrypted` enum doc, and two inline comments to state that deltas are published in plaintext and group membership is not enforced today.
- `src/kv/delta.rs`: Updates one inline comment to remove the implication that anti-entropy merges are safe for `Encrypted` stores due to MLS group auth, replacing it with the accurate reserved-shape description.

<h3>Confidence Score: 5/5</h3>

Safe to merge — exclusively comment and doc changes with zero behavioral impact.

Every changed line is a comment or doc string. The updated text now correctly tells callers that `AccessPolicy::Encrypted` publishes plaintext deltas and allows any writer through, which matches the actual code paths. No logic, data structures, or public APIs are touched.

No files require special attention.

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| src/kv/store.rs | Three comment blocks updated to remove incorrect claims that Encrypted stores encrypt deltas or enforce MLS group membership; behavior and API unchanged. |
| src/kv/delta.rs | One inline comment corrected to drop the false statement that MLS group membership provides authorization for anti-entropy merges; no code changes. |

</details>

</details>

<h3>Flowchart</h3>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[Incoming Delta / Write] --> B{Writer identity present?}
    B -- Yes --> C{Check AccessPolicy}
    B -- No --> D{AccessPolicy::Encrypted?}
    D -- Yes --> E[OK — reserved permissive path\nno group membership check today]
    D -- No --> F[Reject — Signed or Allowlisted\nrequire writer identity]
    C --> G{Signed}
[...]

PR #88, opened 2026-05-21
docs: propose encrypted KvStore design

JimCollinson:
## Summary

Adds a design-only proposal for encrypted group-scoped KvStore sync.

This is not an implementation PR. It is intended as a maintainer-reviewable design prompt before anyone proceeds with code.

## Why

x0x currently has:

- replicated KvStore state over gossip;
- named-group secure content via the secure plane;
- agent signing;
- group rekey semantics.

The missing product primitive is a KvStore scoped to a secure named group where store deltas/checkpoints are encrypted, author-bound, and usable by daemon/API clients without application-side encryption wrappers.

## Review focus

This proposal intentionally highlights the decisions maintainers need to own:

- Which secure backend v1 should target: named-group GSS, current MLS/TreeKEM, or a `SecureContext` abstraction supporting both.
- Whether sign-then-encrypt is the right author-binding model.
- AAD shape: domain/version, `group_id`, `store_id`, epoch; `state_hash` left as an explicit maintainer choice.
- Rekey/checkpoint semantics, including partitioned concurrent rekeys.
- Current-state bootstrap for late joiners.
- Forward-compatible write authorization beyond "any active member may write."
- API shape, with `POST /groups/:id/stores` proposed as the primary route.

## Non-goals

- No implementation.
- No new cryptography.
- No claim that GSS is full MLS TreeKEM.
- No requirement that late joiners receive historical epoch keys.
- No in-place migration of plaintext stores to encrypted stores.

## Relationship to PR #87

PR #87 is the guardrail/doc clarification that `AccessPolicy::Encrypted` is not wired to encrypted KvStore sync today.

This PR is the follow-up design proposal for what encrypted KvStore sync should become.

## Verification

Docs-only change. No Rust files touched, so cargo checks were not run.

<h3>Greptile Summary</h3>

This PR adds a 611-line design-only proposal (`docs/design/encrypted-kvstore.md`) for an encrypted, group-scoped KvStore. No Rust code is changed; the document is intended for maintainer review before any implementation begins.

- **Core proposal:** wrap `KvStoreDelta` and checkpoint publications in a sign-then-encrypt envelope backed by a `SecureContext` abstraction (GSS or MLS/TreeKEM), binding records to `group_id`, `store_id`, and epoch via AEAD AAD and an inner ML-DSA-65 signature.
- **Explicit open questions:** which secure backend v1 targets, who publishes post-rekey checkpoints, and whether write policy beyond active membership is in v1 scope — all correctly flagged as maintainer decisions rather than resolved in the doc.

Safe to merge as a discussion artifact; the document is documentation-only and touches no Rust source files.

The proposal is thorough and well-reasoned, but one design gap — the author-ID-to-public-key binding mechanism — leaves implementers without a concrete, unambiguous verification step in the receive flow. Without resolving whether AgentId is derived from the public key or requires a registry lookup, an implementation could inadvertently allow a group member to spoof another member's authorship. Two smaller items (epoch leakage omitted from the does-not-provide list, and vague nonce birthday-bound guidance) are clarifications that should be addressed before the implementation PR but are not blocking for the design review itself.

docs/design/encrypted-kvstore.md — specifically the receive-flow author-ID verification step and the AEAD nonce guidance section.

| Filename | Overview |
|----------|----------|
| docs/design/encrypted-kvstore.md | New 611-line design proposal for encrypted group-scoped KvStore. Well-structured with clear requirements, user stories, wire-format sketches, rekey semantics, and a test plan. Three items need attention before implementation: author-ID-to-public-key binding mechanism is underspecified in the receive flow, the plaintext epoch field metadata leakage is missing from the security properties list, and nonce guidance ties safety to an undefined traffic volume threshold. |

</details>

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Author as Author Agent
    participant SC as SecureContext
    participant Gossip as Gossip Topic
    participant Receiver as Receiver Agent
    participant KvStore as Local KvStore

    Note over Author,KvStore: Publish Flow
    Author->>Author: Build KvStoreDelta
    Author->>Author: Sign with ML-DSA-65 key
    Author->>SC: seal(SignedRecord, AAD)
    SC-->>Author: EncryptedKvStoreRecordV1
    Author->>Gossip: publish envelope

    Note over Author,KvStore: Receive Flow
    Gossip->>Receiver: EncryptedKvStoreRecordV1
    Receiver->>SC: open(ciphertext, AAD, epoch)
    SC-->>Receiver: SignedRecord
    Receiver->>Receiver: Verify author_id, pubkey, sig
    Receiver->>Receiver: Check write authorization
    alt Valid
        Receiver->>KvStore: merge delta or checkpoint
    else Invalid
        Receiver->>Receiver: Drop record
    end
```

`````markdown

---

### Issue 1 of 3
docs/design/encrypted-kvstore.md:338-342
**Author-ID-to-public-key binding leaves an exploitable gap**

The document requires that "a record signed by one agent but claiming another agent id must be rejected," but then gives two non-equivalent options: the id is "derivable from, or verified against, the included public key." Only the first option — where `author_id` is a deterministic function of the public key (e.g. a hash) — closes the gap in the receive flow as written. The second option ("verified against") requires an out-of-band registry lookup that is not mentioned anywhere in the receive flow's 7 steps. Without specifying which path implementers must take, a valid group member can craft an inner record that includes their own key and signature but claims a different member's `author_id`. The signature check in step 4 passes (key matches signature), but authorship attribution is wrong. The design should either (a) state that `AgentId` is deterministically derived from the ML-DSA-65 public key, or (b) add an explicit "look up the registered public key for `author_id` and compare against the included key" step in the receive flow.

### Issue 2 of 3
docs/design/encrypted-kvstore.md:454-465
**Epoch leakage from plaintext envelope not acknowledged in Security Properties**

`EncryptedKvStoreRecordV1` carries `epoch` in plaintext so receivers know which key to use for decryption — that is a reasonable design choice. However, it means a passive gossip observer who cannot decrypt any content can still observe the exact moment an epoch increment occurs, correlating rekey events (i.e., bans or removes) in near-real time. The "does not provide" list in the Security Properties section currently covers traffic timing and topic membership, but does not mention that epoch-boundary events are observable. Since this is an explicit tradeoff rather than an oversight, adding a bullet there — for example, "hiding the timing or frequency of epoch rotations from gossip observers" — keeps the security-properties section honest and avoids surprises for implementers.

### Issue 3 of 3
docs/design/encrypted-kvstore.md:278-286
**ChaCha20-Poly1305 nonce birthday-bound guidance ties safety to an undefined traffic concept**

The document says a fresh random 96-bit nonce per record is "acceptable at expected application-state volumes," but "expected application-state volumes" is never defined. The birthday bound for a 96-bit nonce space puts the 2^(−32) collision probability threshold at roughly 2^32 ≈ 4 billion records *per epoch key*. For long-lived epochs (no bans/removes), even moderate-write stores could approach that limit over time. Framing XChaCha20-Poly1305 (192-bit nonce) as only needed by "high-churn stores" may mislead implementers into treating it as an edge-case optimization rather than a sensible default for stores where epoch rotation is infrequent. Consider either quantifying the "acceptable volume" threshold or recommending XChaCha20-Poly1305 as the preferred default, with ChaCha20-Poly1305 as the fallback for constrained environments.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["docs: propose encrypted KvStore design"](https://github.com/saorsa-labs/x0x/commit/aed38223785f0a5d32d584fc93a3dbd76de7d39f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=33270818)</sub>


dirvine (commented):
## Review: PR #88 — `docs: propose encrypted KvStore design`

**Verdict: Strong design proposal. Agree with the three Greptile issues. Add one more architectural concern. Should merge as discussion artifact.**

### Strengths
- **Decisions-first table** at the top is excellent — makes maintainer review efficient.
- **Six open questions** correctly identified and scoped (partitioned rekey, checkpoint model, API shape, write-auth v1 scope, grace period).
- **Sign-then-encrypt** is the right choice. GSS is a shared secret; AEAD alone gives no author attribution. Inner ML-DSA-65 signatures inside ciphertext is precisely correct.
- **AAD binding** (domain/version, group_id, store_id, epoch) without state_hash is the right pragmatic choice for v1.
- **SecureContext abstraction** is necessary, not premature — the doc correctly identifies that both GSS and saorsa-mls/TreeKEM are already in the tree. Building the store against either one directly would mean rewriting when the backend shifts.
- **API shape** (POST /groups/:id/stores) is cleaner than optional fields on flat /stores.
- **Forward-compatible write auth** with preserved author ID for later policy layer is exactly right.

### Agreed: Greptile issues (3)
1. **Author-ID-to-public-key binding** (lines ~338-342) — needs to be deterministic. AgentId should be derived from the ML-DSA-65 public key (e.g. BLAKE3 hash), not just "verified against" via registry lookup. Otherwise a valid member can forge another member's author_id with their own key.
2. **Epoch leakage** (lines ~454-465) — add "hiding the timing or frequency of epoch rotations from gossip observers" to the "does not provide" list. Current document covers timing/size but not epoch-boundary observability.
3. **Nonce birthday guidance** (lines ~278-286) — 2^32 records per key before 2^(-32) collision probability. For infrequent-rekey stores this is reachable. Recommend XChaCha20-Poly1305 (192-bit nonce) as the preferred default for all encrypted stores, with ChaCha20-Poly1305 as the documented fallback with an explicit counter-based nonce strategy.

### Additional architectural concern: checkpoint publication ownership
This is the single most sourceable bug vector. If "any member may publish" is the checkpoint rule, a member nearing removal could pre-publish a malicious checkpoint:
- They know the epoch is about to bump (bans are visible).
- They have the new shared secret (still in the group).
- They publish a corrupt/stale checkpoint encrypted at the new epoch.
- Other members see valid-AEAD, valid-epoch content and apply it.

Mitigations to add to the doc:
1. Only the removal-initiating admin/authority publishes the post-rekey checkpoint.
2. The checkpoint must be co-signed by at least two members (quorum).
3. Receivers accept the checkpoint with the highest checkpoint_sequence from a non-removed authority.
4. The checkpoint includes the list of removed members (from the GroupStateCommit), and receivers verify their own membership.

### Summary
Well-structured, honest design document. The decisions table makes maintainer review efficient. The three Greptile issues and the checkpoint-authority concern should be resolved before implementation. **Recommend merging the doc as a discussion artifact.**

No Rust files touched — zero risk.

JimCollinson:
Ok, i've made some amends to this PR inline with your comments @dirvine so they are addressed before the deisgn proposal is merged

dirvine (commented):
## Review: PR #88 — docs: propose encrypted KvStore design

**Verdict: All concerns resolved. Safe to merge as discussion artifact.**

### Amendments review

Jim's second commit (a9eba9d5) directly addresses every concern raised in the initial review:

**Issue 1 — Author-ID-to-public-key binding (Greptile P1, David agreed):** Fully resolved. The ambiguous "derivable from, or verified against" language is replaced with a deterministic requirement: AgentId must be derived from the ML-DSA-65 public key via AgentId::from_public_key. The receive flow now has an explicit derive-and-compare step before signature verification, closing the spoofing gap entirely.

**Issue 2 — Epoch leakage (Greptile P2):** Resolved. Added "hiding the timing or frequency of epoch rotations from gossip observers" to the Security Properties "does not provide" section, with the plaintext-epoch justification clearly stated.

**Issue 3 — Nonce birthday guidance (Greptile P3):** Resolved. The vague "acceptable at expected volumes" language is replaced with XChaCha20-Poly1305 (192-bit nonce) as the preferred default, with explicit birthday-bound math (2^32 records, 2^-32 collision risk) and a persisted-counter fallback strategy for ChaCha20-Poly1305.

**Issue 4 — Checkpoint publication authority (David additional):** Resolved. The section is substantially rewritten. The permissive "any member may publish" option is removed; the document now explicitly flags the removed-member pre-publish attack vector and provides four concrete mitigations (admin-only, quorum, highest-sequence-from-authority, GroupStateCommit inclusion). The framing as a maintainer design-session topic is correct.

### Summary

The amendments are thorough and resolve every raised concern. No Rust files touched. Recommend merging.

dirvine:
Adding the at-rest / local-storage discussion points from review, now recorded as **[ADR 0015: No App-Layer At-Rest Encryption or Secondary Passwords](https://github.com/saorsa-labs/x0x/blob/main/docs/adr/0015-no-app-layer-at-rest-encryption.md)** (Accepted 2026-06-10). This anchors the scope-out this design already states ("Do not solve local at-rest encryption in this document") so it doesn't get re-litigated during implementation:

- **No secondary passwords, ever.** x0xd is an unattended daemon (boot start, self-upgrade restarts, headless VPS). A passphrase-protected local store either breaks unattended startup or the key sits on disk next to the data — no real protection, pure UX cost.
- **OS user isolation is the boundary.** Once an attacker executes code in the user's session they can read decrypted state from daemon memory or just query the localhost REST API — app-layer at-rest encryption is moot against the attacker it implies protection from. If the home dir is compromised, *all* the user's local data is exposed, not just x0x's.
- **Stolen/disposed hardware is full-disk encryption's job** (FileVault default on modern macOS; LUKS/BitLocker elsewhere). Users with stricter needs can mount `~/.x0x` on encrypted volumes.
- **Sanctioned zero-UX follow-up**: best-effort OS-keystore wrapping (Keychain/DPAPI/libsecret) of the ML-DSA identity key files only — no prompt, keys unlock with the login session, plain-file fallback mandatory for headless nodes. This addresses the dumb exfiltration paths (backups, cloud-sync folders, file-grab malware) for the asset that actually matters: identity keys, whose compromise is silent network-wide identity theft rather than a local data leak. (The Signal Desktop key-next-to-database lesson.)
- **Honesty requirement for this design's docs**: a group's E2EE on the gossip path does **not** extend to members' local disks and backups — group confidentiality is bounded by the weakest member's disk hygiene. Worth stating explicitly in the final design, same honesty direction as #87 (now merged).

One additional review note on the open decision table: the "Secure backend target" question (GSS vs TreeKEM vs abstraction) is largely answered by [ADR 0012](https://github.com/saorsa-labs/x0x/blob/main/docs/adr/0012-treekem-default-secure-groups.md) — new private `MlsEncrypted` groups are TreeKEM-by-default, with GSS only for grandfathered groups and public encrypted presets. A `SecureContext` v1 should bind to the TreeKEM plane for private groups.

PR #100, opened 2026-06-09
feat(daemon): app-integration support — per-recipient DM inbox, gossip-DM fallback listeners, fresh-boot DM delivery fixes

dirvine:
## What

Daemon-side support for the communitas Swift/Dioxus apps, **rescued from the x0x working tree** where it sat uncommitted (~1462 lines, one `git checkout` from loss) during the communitas integration review, plus a review fix.

**Draft** because: this was authored outside this session and reviewed here, and it carries **one behavioral decision** (see Blocker) that should be confirmed against the live mesh before merge.

### Change set
- **Gossip-DM: shared-bus → per-recipient inbox topics** (`dm_inbox.rs`, `dm_send.rs`, `gossip/pubsub.rs`). Was: every agent subscribed to one `x0x/dm/v1/bus` and received *every* encrypted DM (O(N) fleet load). Now: sender publishes to `dm_inbox_topic(recipient)`, recipient subscribes to `dm_inbox_topic(self)`; new `subscribe_topic_id`/`publish_topic_id` carry an explicit transport `TopicId`. Legacy bus retained as a rolling-upgrade fallback with a 250ms ACK-then-fallback path. Scalability + privacy win, rolling-upgrade-safe.
- **Cached `/upgrade` response** — `tokio::sync::Mutex` (no poison panic), 6h success / 30min error TTL, anti-hammer for polling clients.
- **Two background gossip-DM fallback listeners** (signed-public-message + KvStore-delta) — reliability backstop when the topic mesh is late/congested.
- **Per-request `require_gossip_ack` override** on `POST /direct/send`.
- **lib.rs**: peer-ID-authenticated local dials (security fix — old code trusted whoever answered a LAN address); discovery-cache merge helpers; `put_with_delta`/`apply_remote_delta` KvStore API.
- **Review fix (this PR, `ed6f845`)**: bound the discovered-agent address list — restore replace-on-fresher (the WIP had regressed to unbounded union → roaming agents accumulated dead endpoints) + the missing unit tests.

## Review verdict: ship-able after one decision

Reviewed by 3 agents + direct review. **Builds clean, clippy `-D warnings` clean, 190/190 DM/gossip/group tests pass.**

**Security — CLEAN.** Both new fallback listeners gate on `verified` (ML-DSA-65 sig + sender-id↔key binding) and then call the **same** `ingest_public_message` / `merge_delta` the gossip path uses → full author-ban + write-policy + per-message signature re-check. No transport-`verified`-gate bypass; a non-member cannot inject deltas. Dual gossip+direct delivery is idempotent (dedup by signature / CRDT merge) — no double-apply.

## 🔴 Blocker (human decision): `require_gossip_ack` default flip

`direct_message_send_config()` dropped its explicit `require_gossip_ack: false`, so generic daemon/UI DMs now inherit the default `true`. The removed comment warned this exact flip causes false 504s ("live meshes deliver the payload while dropping the reverse ACK; surfacing that as 504 makes users resend messages the recipient already displayed"). **The communitas client never sends the field**, so all its UI DMs now require the ACK.

There's a coherent case it's now safe — the *same* change set hardened the ACK path (per-recipient inbox + 250ms ACK-then-legacy-fallback). But it reverses a fix for a real observed problem. **Decide one of:**
- (a) validate on the live mesh that ACKs now arrive reliably (no false 504s), or
- (b) have the communitas client pass `require_gossip_ack: false` for fire-and-forget UI sends.

## Remaining (non-blocking)
- LOW: `public_messages` map unbounded in *number of groups* (per-group history is capped); listener tasks lack a graceful-shutdown hook; KV errors tunneled through `IdentityError::Storage`; `connect_to_machine` probes LAN regardless of `should_attempt_direct()`.

## Not included
- Client-side app blockers from the communitas review (WS reconnect in both apps; Dioxus macOS receive gate) — those live in the communitas repo, not here.

<h3>Greptile Summary</h3>

This PR adds daemon-side app-integration support: per-recipient gossip-DM inbox topics (replacing the shared bus), two background fallback listeners for group-public-message and KvStore-delta delivery over DM, a cached `/upgrade` response, peer-ID-authenticated LAN dials, a `upsert_discovered_agent` helper that bounds the address list, and a `CapabilityStore` timestamp-ordering fix to prevent stale pending adverts from poisoning the gossip-ready entry.

- **DM inbox topology**: senders now publish to `dm_inbox_topic(recipient)` with a 250 ms ACK-then-legacy-bus fallback; the legacy shared bus is retained as a rolling-upgrade backstop. `RecentDeliveryCache::insert` returns a bool (atomic slot-claim) to prevent dual-path double-delivery.
- **`direct_message_send_config` behavioral change**: `require_gossip_ack` now inherits the default `true` (previously explicit `false`), and `stop_fallback_on_raw_error` is flipped to `false` for file transfers — both are deliberate reversals of prior design decisions that need live-mesh validation before merge.
- **Security fix**: `connect_to_agent` LAN dials now authenticate the peer ID before trusting the connection; the unauthenticated path is retained as a fallback only when no `machine_id` hint is available.

<h3>Confidence Score: 3/5</h3>

Not safe to merge until the `require_gossip_ack` default change is validated or the communitas client is updated to pass the field explicitly.

The core DM inbox re-architecture (per-recipient topics, dedupe, legacy fallback, capability-advert poisoning fix, peer-ID-authenticated dials) is solid and well-tested. The unresolved concern is `direct_message_send_config`: dropping `require_gossip_ack: false` changes how every communitas UI DM behaves — the client never sends the field, so all its sends now wait for a gossip ACK that the PR description warns can be dropped on live meshes without the payload being lost, surfacing as a false 504. The `stop_fallback_on_raw_error` flip for file transfers also reverses an explicit prior decision (gossip fanout breaking chunk throughput) without fully addressing the original concern.

src/bin/x0xd.rs — `direct_message_send_config` and `file_transfer_send_config` contain the two behavioral reversals that need a decision before merge.

| Filename | Overview |
|----------|----------|
| src/bin/x0xd.rs | Major daemon changes: two new background gossip-DM fallback listener tasks, per-request `require_gossip_ack` override, cached `/upgrade` response (tokio Mutex, 6h/30min TTL), `--disable-peer-cache` flag, and `direct_message_send_config` behavioral flip (`require_gossip_ack` now defaults true, `stop_fallback_on_raw_error` flipped for file transfers). |
| src/dm_inbox.rs | New per-recipient inbox subscription service. Primary path uses `subscribe_topic_id` with domain-separated TopicId; legacy bus retained as rolling-upgrade fallback. Atomic dedupe claim in `cache.insert` prevents dual-path double-delivery. Well-tested with 20+ unit tests. |
| src/dm_send.rs | Sender-side gossip DM path updated to publish to per-recipient `publish_topic_id` topics with 250ms ACK-then-legacy-bus fallback. X0X-0041 lifecycle-hint backoff short-circuit well-tested. |
| src/lib.rs | New `upsert_discovered_agent` replaces unbounded union with replace-on-fresher semantics for addresses. Peer-ID-authenticated LAN dials in `connect_to_agent` (security fix). New `connect_to_machine` LAN probes don't guard on `should_attempt_direct()`, causing dial attempts to relay-only peers. |
| src/gossip/pubsub.rs | New `subscribe_topic_id` and `publish_topic_id` overloads that accept an explicit transport TopicId. Existing `subscribe`/`publish` now delegate to these, no behavioral change for existing callers. |
| src/dm_capability.rs | CapabilityStore now orders adverts by `created_at_unix_ms` to prevent stale pending adverts from clobbering gossip-ready ones; includes tests covering the out-of-order delivery scenario. |
| src/dm.rs | `RecentDeliveryCache::insert` now returns `bool` acting as atomic slot-claim to prevent dual-path (primary inbox + legacy bus) double-delivery; covered by new test. |
| src/direct.rs | Adds `incoming_typed_route_dropped` diagnostic counter for best-effort typed-route fallback drops; new `serde(default)` attribute ensures backward-compatible JSON deserialization. |
| src/dm_capability_service.rs | Publisher now guards against broadcasting not-yet-usable (pending) adverts, restarting burst on capability upgrade. Clean fix for capability-advert poisoning described in the changelog. |

</details>

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Sender
    participant PubSub
    participant RecipientInbox as Recipient Inbox (new)
    participant LegacyBus as Legacy Bus (fallback)
    participant InboxService as DmInboxService
    participant DM as DirectMessaging

    Sender->>PubSub: publish_topic_id(dm_inbox_topic(recipient), wire)
    PubSub-->>RecipientInbox: deliver envelope

    alt ACK within 250ms
        InboxService-->>Sender: ACK via publish_topic_id
    else no ACK (rolling upgrade / congestion)
        Sender->>PubSub: publish(DM_BUS_TOPIC, wire)
        PubSub-->>LegacyBus: deliver envelope
    end

    InboxService->>InboxService: verify sig, decrypt, dedupe (atomic cache.insert)
    InboxService->>DM: handle_incoming (or typed route)
    InboxService-->>Sender: publish_ack via inbox_topic OR legacy_bus

    Note over InboxService: Dual-path dedup: cache.insert returns false for loser task
```

<details open><summary><h3>Comments Outside Diff (2)</h3></summary>

1. `src/bin/x0xd.rs`, line 2619-2631 ([link](https://github.com/saorsa-labs/x0x/blob/03197b1dc904e765992cbe7665ae9d4a936d635a/src/bin/x0xd.rs#L2619-L2631)) 

   <a href="#">P1</a> **`require_gossip_ack` default flip affects all communitas UI DMs**

   `direct_message_send_config()` dropped `require_gossip_ack: false`, so the field now inherits `DmSendConfig::default()` which is `true`. The PR description acknowledges this as a 🔴 Blocker: the communitas client never passes `require_gossip_ack` in its `POST /direct/send` body, so every UI DM now blocks until the recipient ACKs via the gossip inbox. On a live mesh where the reverse ACK can be dropped (especially in the first seconds after boot or during a rolling upgrade), this surfaces as a 504 to the sender — who then resends a message the recipient already displayed. The prior comment capturing this exact failure scenario was removed. Even with the hardened per-recipient inbox and 250 ms ACK-then-legacy-fallback, the original observed problem (ACK drop ≠ payload loss) is not eliminated — it just becomes less frequent. A decision is needed before merge: either validate ACKs arrive reliably on the live mesh, or add `require_gossip_ack: false` in the communitas client for fire-and-forget UI sends.

   <details><summary>Prompt To Fix With AI</summary>

   `````markdown
   This is a comment left during a code review.
   Path: src/bin/x0xd.rs
   Line: 2619-2631

   Comment:
   **`require_gossip_ack` default flip affects all communitas UI DMs**

   `direct_message_send_config()` dropped `require_gossip_ack: false`, so the field now inherits `DmSendConfig::default()` which is `true`. The PR description acknowledges this as a 🔴 Blocker: the communitas client never passes `require_gossip_ack` in its `POST /direct/send` body, so every UI DM now blocks until the recipient ACKs via the gossip inbox. On a live mesh where the reverse ACK can be dropped (especially in the first seconds after boot or during a rolling upgrade), this surfaces as a 504 to the sender — who then resends a message the recipient already displayed. The prior comment capturing this exact failure scenario was removed. Even with the hardened per-recipient inbox and 250 ms ACK-then-legacy-fallback, the original observed problem (ACK drop ≠ payload loss) is not eliminated — it just becomes less frequent. A decision is needed before merge: either validate ACKs arrive reliably on the live mesh, or add `require_gossip_ack: false` in the communitas client for fire-and-forget UI sends.

   How can I resolve this? If you propose a fix, please make it concise.
   `````
   </details>

2. `src/lib.rs`, line 3063-3072 ([link](https://github.com/saorsa-labs/x0x/blob/03197b1dc904e765992cbe7665ae9d4a936d635a/src/lib.rs#L3063-L3072)) 

   <a href="#">P2</a> **`connect_to_machine` LAN probes ignore `should_attempt_direct()`**

   The new local-probe loop in `connect_to_machine` unconditionally attempts two peer-authenticated dials per LAN address (3 s each: `connect_peer_with_addrs` + `connect_addr` fallback) without checking `info.should_attempt_direct()`. For relay-only peers (`can_receive_direct = false`) this triggers up to 6 seconds of dial timeouts per LAN address before the code continues to the NAT traversal step. The PR description notes this as a known non-blocking item ("LOW: `connect_to_machine` probes LAN regardless of `should_attempt_direct()`"); adding the guard would align the new path with the existing `connect_to_agent` behavior.

   <details><summary>Prompt To Fix With AI</summary>

   `````markdown
   This is a comment left during a code review.
   Path: src/lib.rs
   Line: 3063-3072

   Comment:
   **`connect_to_machine` LAN probes ignore `should_attempt_direct()`**

   The new local-probe loop in `connect_to_machine` unconditionally attempts two peer-authenticated dials per LAN address (3 s each: `connect_peer_with_addrs` + `connect_addr` fallback) without checking `info.should_attempt_direct()`. For relay-only peers (`can_receive_direct = false`) this triggers up to 6 seconds of dial timeouts per LAN address before the code continues to the NAT traversal step. The PR description notes this as a known non-blocking item ("LOW: `connect_to_machine` probes LAN regardless of `should_attempt_direct()`"); adding the guard would align the new path with the existing `connect_to_agent` behavior.

   How can I resolve this? If you propose a fix, please make it concise.
   `````
   </details>

</details>

`````markdown

---

### Issue 1 of 3
src/bin/x0xd.rs:2619-2631
**`require_gossip_ack` default flip affects all communitas UI DMs**

`direct_message_send_config()` dropped `require_gossip_ack: false`, so the field now inherits `DmSendConfig::default()` which is `true`. The PR description acknowledges this as a 🔴 Blocker: the communitas client never passes `require_gossip_ack` in its `POST /direct/send` body, so every UI DM now blocks until the recipient ACKs via the gossip inbox. On a live mesh where the reverse ACK can be dropped (especially in the first seconds after boot or during a rolling upgrade), this surfaces as a 504 to the sender — who then resends a message the recipient already displayed. The prior comment capturing this exact failure scenario was removed. Even with the hardened per-recipient inbox and 250 ms ACK-then-legacy-fallback, the original observed problem (ACK drop ≠ payload loss) is not eliminated — it just becomes less frequent. A decision is needed before merge: either validate ACKs arrive reliably on the live mesh, or add `require_gossip_ack: false` in the communitas client for fire-and-forget UI sends.

### Issue 2 of 3
src/bin/x0xd.rs:2695-2700
The original comment explicitly called out that gossip fanout for 32 KiB file-transfer chunks breaks the throughput/back-pressure window: `stop_fallback_on_raw_error: true` kept raw failures terminal for that reason. With `false`, a raw receive-ACK failure now triggers a gossip fallback that pushes those chunks through signed/encrypted pubsub fanout. The new comment says `DEFAULT_CHUNK_SIZE is sized to fit the DM envelope cap`, but the fanout cost (fan-out to every subscribed peer, not just the recipient) and back-pressure implications are unchanged. If gossip fallback is intentionally enabled here, the original design concern should be addressed explicitly, or the old guard restored while only changing `raw_quic_receive_ack_timeout`.

```suggestion
fn file_transfer_send_config() -> x0x::dm::DmSendConfig {
    let mut config = x0x::dm::DmSendConfig {
        prefer_raw_quic_if_connected: true,
        stop_fallback_on_raw_error: true,
        ..x0x::dm::DmSendConfig::default()
    };
```

### Issue 3 of 3
src/lib.rs:3063-3072
**`connect_to_machine` LAN probes ignore `should_attempt_direct()`**

The new local-probe loop in `connect_to_machine` unconditionally attempts two peer-authenticated dials per LAN address (3 s each: `connect_peer_with_addrs` + `connect_addr` fallback) without checking `info.should_attempt_direct()`. For relay-only peers (`can_receive_direct = false`) this triggers up to 6 seconds of dial timeouts per LAN address before the code continues to the NAT traversal step. The PR description notes this as a known non-blocking item ("LOW: `connect_to_machine` probes LAN regardless of `should_attempt_direct()`"); adding the guard would align the new path with the existing `connect_to_agent` behavior.


`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(dm): close the fresh-boot DM deliver..."](https://github.com/saorsa-labs/x0x/commit/03197b1dc904e765992cbe7665ae9d4a936d635a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36591409)</sub>

> Greptile also left **1 inline comment** on this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+
+- **Fresh-boot DM delivery black hole (dogfood `group_join` / hop-DM 25s timeouts).** Local 3-daemon soaks failed 10-17% of iterations with `group_join timed out` / `hop DM never echoed back`; a trace-instrumented capture run pinned a chain of three compounding faults, each masked by the previous one:
+  1. **Capability-advert poisoning**: `CapabilityAdvertService` broadcast the startup `pending()` advert (no gossip inbox / KEM key), and `CapabilityStore::insert` was unconditional last-writer-wins — epidemic broadcast delivers out of order, so a stale pending advert could clobber the gossip-ready one and leave senders on `advert_cache_unusable`. The store now orders adverts by their signed `created_at_unix_ms` (stale ignored; a genuinely fresher downgrade still wins), and the publisher never broadcasts a not-yet-usable advert (absence already means "use the raw fallback").
+  2. **Fire-and-forget raw-QUIC loss**: the daemon's generic DM config sent raw-path messages without ant-quic's receive-pipeline ACK — a send into a connection being superseded during boot churn returned `Ok` (dur_ms=0) while the bytes were lost, so the retry machinery never fired and the recipient's app never saw the message. `direct_message_send_config()` now sets `raw_quic_receive_ack_timeout = 8s` (matching the named-group config), so loss fails the attempt and the existing retry re-sends.
+  3. **Zombie-connection retry pinning**: with loss now detected, retries still resolved `cached_connected` onto the same dead connection (`is_connected` stays true for a zombie whose remote endpoint vanished without a lifecycle event) and burned ~14s per attempt until the caller's deadline. An ACKed-path send failure on a still-connected peer now tears the connection down so the next attempt takes the X0X-0031/0033 send-readiness repair (fresh dial).
+
+  Validated: the capture harness went from failing within 12-27 iterations to 40/40 clean; `tests/pr99_local_soak.sh` restored to baseline (see soak artefacts under `proofs/x0x-gui-full-dogfood/`).
+
 ### Added
 
 - **`welcome.trace` diagnostics for the TreeKEM Welcome-blob transfer.** Trace-only (no behavior change): `target=welcome.trace` debug stages on both sides of the chunked Welcome pull — anchor `offer_sent` / `chunk_sent` / `chunk_ack_recv` / `final_ack_{ok,failed}`; receiver `chunk_recv` / `chunk_recv_no_pending` / `chunk_ack_sent`. Enable with `RUST_LOG=warn,welcome.trace=debug` to locate exactly where a Welcome transfer stalls under churn.
--- a/src/bin/x0xd.rs
+++ b/src/bin/x0xd.rs
@@ -2625,9 +2625,13 @@ fn named_group_direct_delivery_config() -> x0x::dm::DmSendConfig {
     // The gossip-inbox DM path verifies the signed DM envelope and marks the
     // bridged direct message verified. Raw QUIC can only mark messages
     // verified when the receiver already has a fresh AgentId -> MachineId
-    // binding, so it must remain a fallback rather than preempting gossip.
+    // binding, so keep it as the fallback for peers whose gossip-inbox
+    // capability advert has not converged yet. Terminal signed commits (for
+    // example creator delete) are self-authenticating and explicitly re-check
+    // authority on apply, so dropping the raw fallback can strand members after
+    // their metadata listener exits.
     let mut config = direct_message_send_config();
-    config.require_gossip = true;
+    config.raw_quic_receive_ack_timeout = Some(Duration::from_secs(8));
     config.require_gossip_ack = true;
     config
 }
@@ -20624,6 +20628,10 @@ mod tests {
         assert!(!config.prefer_raw_quic_if_connected);
         assert!(!config.require_gossip);
         assert!(!config.stop_fallback_on_raw_error);
+        assert_eq!(
+            config.raw_quic_receive_ack_timeout,
+            Some(Duration::from_secs(8))
+        );
     }
 
     #[test]
--- a/src/bin/x0xd.rs
+++ b/src/bin/x0xd.rs
@@ -2614,8 +2614,17 @@ fn direct_message_send_config() -> x0x::dm::DmSendConfig {
     // Generic daemon/UI DMs should only return success after the inbox path
     // observes the recipient ACK. Callers that intentionally want
     // fire-and-forget gossip can pass `require_gossip_ack: false`.
+    //
+    // The raw-QUIC fallback (taken whenever the recipient's gossip-inbox
+    // capability advert has not converged yet — always the case in the first
+    // seconds after boot) must use ant-quic's receive-pipeline ACK. A
+    // fire-and-forget raw send into a connection that is being superseded
+    // reports Ok while the bytes are lost, the retry machinery never fires,
+    // and the recipient's app never sees the message (the dogfood
+    // group_join / hop-DM 25s-timeout black hole).
     x0x::dm::DmSendConfig {
         timeout_per_attempt: Duration::from_secs(8),
+        raw_quic_receive_ack_timeout: Some(Duration::from_secs(8)),
         ..x0x::dm::DmSendConfig::default()
     }
 }
@@ -4510,7 +4519,12 @@ async fn import_agent_card(
     if machine_id_bytes != [0u8; 32] {
         if let Some(caps) = card.dm_capabilities.clone() {
             if caps.gossip_inbox && !caps.kem_public_key.is_empty() {
-                capability_store.insert(agent_id, x0x::identity::MachineId(machine_id_bytes), caps);
+                capability_store.insert(
+                    agent_id,
+                    x0x::identity::MachineId(machine_id_bytes),
+                    caps,
+                    x0x::dm_capability::now_unix_ms(),
+                );
                 inserted_dm_capability = true;
             }
         }
@@ -19964,7 +19978,15 @@ mod tests {
 
     #[test]
     fn direct_message_send_config_requires_gossip_ack_by_default() {
-        assert!(direct_message_send_config().require_gossip_ack);
+        let config = direct_message_send_config();
+        assert!(config.require_gossip_ack);
+        // Raw-QUIC fallback must be loss-detecting (receive-pipeline ACK), or
+        // a send into a superseded connection reports Ok, the retry never
+        // fires, and the recipient's app never sees the message.
+        assert_eq!(
+            config.raw_quic_receive_ack_timeout,
+            Some(Duration::from_secs(8))
+        );
     }
 
     #[test]
--- a/src/bin/x0xd.rs
+++ b/src/bin/x0xd.rs
@@ -28,7 +28,7 @@ use axum::body::Bytes;
 use axum::extract::{Path, Query, State};
 use axum::http::{HeaderMap, HeaderValue, StatusCode};
 use axum::response::sse::{Event, Sse};
-use axum::response::IntoResponse;
+use axum::response::{IntoResponse, Response};
 use axum::routing::{delete, get, patch, post, put};
 use axum::{Json, Router};
 use base64::Engine;
@@ -522,6 +522,8 @@ struct AppState {
     shutdown_notify: watch::Sender<bool>,
     /// Update configuration honored by manual API-triggered update checks.
     update_config: DaemonUpdateConfig,
+    /// Cached `/upgrade` response so polling clients cannot hammer GitHub.
+    upgrade_check_cache: Mutex<Option<CachedUpgradeCheck>>,
     /// Serializes all destructive binary replacement attempts.
     upgrade_apply_lock: Arc<Mutex<()>>,
     /// API bearer token for authenticating local clients.
@@ -536,6 +538,17 @@ struct AppState {
 /// is populated from untrusted discovery surfaces, so it must not grow without
 /// bound even if every incoming card is syntactically valid.
 const GROUP_CARD_CACHE_CAP: usize = 8_192;
+
+const UPGRADE_CHECK_CACHE_TTL: Duration = Duration::from_secs(6 * 60 * 60);
+const UPGRADE_CHECK_ERROR_CACHE_TTL: Duration = Duration::from_secs(30 * 60);
+
+#[derive(Clone)]
+struct CachedUpgradeCheck {
+    checked_at: Instant,
+    status: StatusCode,
+    body: serde_json::Value,
+    ttl: Duration,
+}
 
 fn group_card_expiry_millis(card: &x0x::groups::GroupCard) -> u64 {
     if card.expires_at > card.issued_at {
@@ -983,6 +996,11 @@ struct DirectSendRequest {
     /// gossip DM capability.
     #[serde(default)]
     require_gossip: bool,
+    /// If set, override whether gossip-inbox sends wait for the recipient's
+    /// inbox ACK before returning success. When omitted, the daemon default is
+    /// used.
+    #[serde(default)]
+    require_gossip_ack: Option<bool>,
     /// Optional opt-in: after the DM path accepts the message, probe the
     /// recipient's ant-quic receive pipeline for liveness with this timeout.
     /// This does not force the message itself onto raw-QUIC receive-ACK.
@@ -1161,6 +1179,7 @@ async fn main() -> Result<()> {
         println!("    --name <NAME>                   Instance name for multi-instance support");
         println!("    --api-port <PORT>               Override API server port");
         println!("    --no-hard-coded-bootstrap       Skip configured bootstrap peers");
+        println!("    --disable-peer-cache            Do not load or save cached peers");
         println!("    --exec-acl <PATH>               Override default exec ACL path");
         println!("    --check                         Check configuration and exit");
         println!("    --check-updates       Check for updates and exit");
@@ -1212,6 +1231,7 @@ async fn main() -> Result<()> {
     // where operator policy forbids unsolicited router port mappings. This
     // overrides the daemon config's `port_mapping_enabled` field.
     let cli_no_port_mapping = args.contains(&"--no-port-mapping".to_string());
+    let cli_disable_peer_cache = args.contains(&"--disable-peer-cache".to_string());
 
     // Parse --api-port for overriding the API server port
     let api_port_override = if let Some(idx) = args.iter().position(|a| a == "--api-port") {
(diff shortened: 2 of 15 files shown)


dirvine:
## Multi-agent deep review complete + 2 fixes applied (`1094802`)

Ran a 22-agent deep review (5 dimensions × adversarial per-finding verification). **Refuted 5 of 17 candidates — including the two scariest security findings**, confirming the new gossip-DM fallback listeners genuinely re-verify ML-DSA signatures + enforce the same authorization as the gossip path (no `verified`-gate bypass, no non-member injection). Net: **1 MEDIUM + 11 LOW** confirmed; no CRITICAL/HIGH survived.

### Fixed in `1094802`
- **MEDIUM — head-of-line backpressure → DM ACK stalls.** `route_typed_payload` did `send().await` on the bounded(1024) typed-route channels, awaited inline in the DM-inbox loop that also publishes ACKs; a slow/contended group-public or kv-delta consumer could stall unrelated DM ACKs (→ 504s under `require_gossip_ack=true`). Now uses `try_send` (these are redundant fallback paths; primary fan-out is per-group/store pubsub), dropping on a full channel and counting it via a new `incoming_typed_route_dropped` counter in `/diagnostics/dm`.
- **LOW — dual-path double-delivery.** Primary inbox + legacy bus could both deliver the same DM (both miss the `handle_incoming` lookup before either delivers). `RecentDeliveryCache::insert` now returns a claim bool; `handle_payload` claims the dedupe slot atomically **after decrypt, before delivery** → exactly one task delivers, the loser re-ACKs.

Validated: fmt + clippy `-D warnings` clean; 275/275 dm/inbox/direct/gossip tests (incl. a new claim-once test).

### `require_gossip_ack` blocker — de-risked
The MEDIUM was the concrete mechanism that would have turned the `false→true` default flip into false 504s (delivered payload, stalled ACK). With the backpressure removed, that vector is closed and `require_gossip_ack=true` is defensible. **Recommended final gate: one live-mesh DM round-trip** to confirm reverse ACKs arrive reliably end-to-end before un-drafting.

### Remaining LOWs (non-blocking)
Rolling-upgrade: `require_gossip_ack=false` never publishes to the legacy bus (NEW→OLD fire-and-forget undeliverable mid-upgrade); `KvStoreHandle::put` now returns `Err` where it returned `Ok` (fail-loud, document); detached listener tasks don't observe `shutdown_notify`; KvStore writer = DM sender attribution (safe, mirrors gossip); the files-recipient e2e stubs `window.api` (doesn't exercise real `/files/send`).

dirvine:
## Live-mesh DM smoke ✅ (local two-daemon, wip @ `1094802`)

Ran `e2e_dogfood_local.sh` (boots two daemons, full bidirectional DM round-trip via `/direct/send` → now `require_gossip_ack=true`) **4× back-to-back: 19/19 pass, 0 fail, ~1.1s each.**

This confirms the `require_gossip_ack` default flip is safe on a local mesh: the send reliably receives its ACK (a 504 would fail the round-trip), bob receives + echoes `received_dm` exactly once (single delivery), no flakiness. Combined with the backpressure fix (which removed the ACK-stall vector), the false-504 risk is closed for local/LAN.

**Caveat (scope):** loopback ACK latency is near-instant, so this validates the path but doesn't stress cross-region ACK latency — the original 504 issue was observed on the live VPS mesh. A VPS-testnet DM matrix would be the ultimate confirmation before un-drafting for production rollout.

dirvine:
## 1h VPS testnet soak ✅ GO (6-node cross-region mesh, wip @ `1094802`)

Cross-compiled + deployed to all 6 **testnet** instances (`x0xd-testnet.service`, 6483/13600 — prod untouched); deploy gate clean (24/24, Phase-A DM matrix + Phase-B 59/59). Then `launch_soak.py --duration-hours 1 --interval-mins 15 --anchor nyc --gate broad-launch`.

**Soak verdict: GO (4 windows).**

| Window | Phase-A DM matrix | drop_full | scenario_disp_to | pp_to |
|---|---|---|---|---|
| 1 | 30/**29** | 0 | 0 | 374 |
| 2 | 30/**29** | 0 | 0 | 216 |
| 3 | 30/**30** | 0 | 2 | 274 |
| 4 | 30/**30** | 0 | 5 | 315 |

- **Phase-A climbs to 30/30** by windows 3-4 — early misses were mesh-formation (deploy just restarted all daemons); at steady state **all 30 cross-region DM pairs deliver under `require_gossip_ack=true`**. The false-504 concern does **not** materialize.
- `drop_full=0` every window. Per-window 'NO-GO' is the known-tight broad-launch per-peer-timeout bar (`pp_to`), not delivery failure — soak-level verdict is GO.

**`/diagnostics/dm` on nyc (pre → post soak):**
- `incoming_typed_route_dropped`: 0 → **0** (try_send never dropped over the hour)
- `incoming_envelopes_total` == `incoming_delivered_to_subscribe`: 142/142 → **923/923** — delivered tracks received exactly across 923 msgs ⇒ **zero double-delivery** on the real mesh.
- `decode_failed`/`signature_failed`: 0/0
- outgoing fail: 30% (formation) → **~0.8% incremental during soak** (612 sent / 5 failed) — residual = normal cross-region transient.

### Verdict
Both fixes validated on the real cross-region mesh, and the `require_gossip_ack=true` default is confirmed safe (no false-504 storm; steady-state DM delivery 30/30 + ~99.2% send success). **This was the production gate the PR was waiting on — recommend un-drafting / ready for merge** once a reviewer signs off. (Per-peer-timeout bar tightness is a pre-existing harness calibration item, not introduced here.)

dirvine:
## ⚠️ Release blocked — 2 deterministic regressions vs main (found via CI, not flakes)

CI flagged 2 failing tests; I verified they are **deterministic** (fail 4/4 CI + local), **pass on main**, and **bisect to the rescued work (`21aaf0c`), not the review fixes**:

### 1. `test_ttl_expiry_removes_from_presence` — FIXED (`7bdca0b`)
`upsert_discovered_agent` clamped `last_seen` with `.max()`, so the presence/TTL filter could never observe an aged entry. Restored `last_seen = incoming.last_seen` (matches main). Verified pass + 3 upsert tests.

### 2. `named_group_creator_delete_propagates_to_peer` — OPEN (deep TreeKEM convergence regression)
Traced (X0X_LOG_DIR + treekem.trace):
- GroupDeleted **does** reach bob (`apply_metadata_event_entry event=group_deleted verified=true allow_queue=true`) — delivery is fine.
- But bob's **join never converges**: `member_added_state_commit_apply_failed`, `error="stale revision: got 1, have 1"`, `commit_prev_state_hash` ≠ bob's `local_state_hash` → alice/bob diverge on group state-hash.
- So GroupDeleted (built on alice's post-join state) can't match bob's state → **queued, never applied** → bob never sees deletion.

Root cause is in the TreeKEM secure-group state-commit / invite-stub convergence chain (the apply code is unchanged by this PR — the rescued per-recipient-inbox delivery-ordering rewrite perturbed convergence). This is the most delicate area of the codebase and needs the rescue author's context.

**Scope note:** the multi-agent review + 1h testnet soak passed because they exercised DM delivery (works) and public-group flows — neither drove a **secure-group join→delete convergence**, which is what's broken. Recommend secure-group convergence coverage before this merges.

**Status: do not merge/release until #2 is fixed and re-verified.**

PR #102, opened 2026-06-10
Batch 1: fix #101 DM capability black-hole; add user-id inspect, sign domain separation, warn-default logging (#85 #90 #93)

### LINKED ISSUE #93: Add daemonless `x0x user-id inspect [PATH]` for reading and validating user identity files
## Summary

Following PR #92 (`x0x user-id create [PATH]`), it would be useful to
have a sibling daemonless command for reading and validating user
identity files:

- Proposed shape: `x0x user-id inspect [PATH]`
- Default `PATH`: `~/.x0x/user.key`
- Behaviour: read the file, attempt to deserialise as a `UserKeypair`
  using the existing storage primitive, and report.
- Must not require a running `x0xd` daemon. Pure local file operation,
  the symmetric counterpart of `x0x user-id create`.

## Proposed output

**On success** (text mode):

```text
User identity at /home/user/.x0x/user.key:
user_id:    0xabcd1234...
user_words: alice.fox.river.dawn
```

**On success** (JSON mode, `--json`):

```json
{
  "path": "/home/user/.x0x/user.key",
  "user_id": "0xabcd1234...",
  "user_words": "alice.fox.river.dawn",
  "valid": true
}
```

**On failure**: exit non-zero with a clear error message identifying
the file and the reason (file missing, unreadable, malformed,
deserialisation failed, etc.).

`user_words` would be derived using x0x's existing
`four_word_networking::IdentityEncoder` (same code path
`inject_identity_words` already uses for daemon API responses), so no
new cryptographic surface is introduced.

## Motivation

External applications that consume x0x's identity primitives need a
way to validate a `user.key` file without spinning up `x0xd`. Concrete
example: Hightea (a TypeScript application coordinating AI agent
skills) is implementing an `init` flow that creates or reuses
`~/.x0x/user.key`. Per its design, init is a one-shot setup step that
must not start a daemon — daemon startup is deferred to whichever
command first actually needs it (the "lazy daemon startup" pattern).

Today, Hightea has two options for an existing key file at init:

1. Start the daemon to validate the key (walks back the lazy-startup
   design).
2. Trust the file exists and defer cryptographic validation to first
   daemon start. (This is what Hightea will ship with v0.2.)

Option 2 is acceptable but means users can be in a "broken state"
after init (corrupted `user.key`) without discovering it until they
run their first daemon-needing command. A daemonless `inspect` would
let Hightea — and any other agentic consumer of x0x — validate
identity files at the point they are first encountered, without
violating the never-reimplement-x0x rule that keeps cryptographic
responsibility inside x0x.

The same gap will exist for any future tool that wants to display the
four-word form of a UserId from a file (e.g. a `whoami`-style command,
a key-management UI, a diagnostic script) without taking a dependency
on a running daemon.

## Non-blocking

This is a quality-of-life addition for downstream consumers, not a
blocker on any in-flight work. Hightea v0.2 will ship with the
trust-and-defer approach above. We are surfacing the gap now so it
can sit in x0x's backlog and be picked up when convenient.

## Alternatives considered

- **Require consumers to start a daemon to validate keys.** Works
  today but couples identity-file validation to daemon lifecycle, and
  prevents lightweight "is this key OK?" tooling. Also makes every
  agentic consumer's setup flow heavier than it needs to be.
- **Expose the storage primitive as a library API.** Already the case
  for in-process Rust consumers via `x0x::storage::load_user_keypair_from`,
  but non-Rust consumers (Node, Python, etc.) cannot use it directly
  and the never-reimplement rule means they should not reimplement
  bincode parsing locally. A CLI subcommand is the language-agnostic
  interface.
- **Add a `--validate` flag to `x0x user-id create`.** Conflates
  create-and-validate into one command. Cleaner to have create / inspect
  as symmetric siblings.

## Related

- PR #92 (`x0x user-id create [PATH]`) — the create-side counterpart
  this would symmetrise with.
- `inject_identity_words` in `src/cli/commands/identity.rs` — the
  existing daemon-side path that derives `user_words` and would be
  reused here.

## Out of scope

This issue does not request:

- A general "inspect any key file" command — keep the scope tight to
  user identity files for now.
- Editing or rotating user keys — read-only inspection only.
- API endpoint exposure — the daemon already has `/agent` and
  `/agent/user-id`. This is specifically the daemonless gap.

dirvine:
## Summary

First batch of the open-issue cleanup (#101, #93, #90, #85). All four verified locally: clippy `-D warnings` clean, `cargo doc -D warnings` clean, full workspace nextest **1680/1680**, plus daemon-backed sign tests 7/7.

### #101 — DM capability upgrade lost at startup (cross-NAT DM black-hole)
`start_dm_inbox` published the capability upgrade with `watch::Sender::send`, which **drops the value when no receiver is subscribed**. x0xd calls `start_dm_inbox` before `start_capability_advert_service` (x0xd.rs:2664-2668), so the upgrade vanished; peers cached `gossip_inbox=false` forever and used the raw-QUIC fallback that black-holes across NAT. Fixed with `send_replace`. The new regression test `dm_inbox_capability_upgrade_visible_to_late_subscriber` was verified to **fail on the pre-fix code** and pass post-fix.

### #93 — daemonless `x0x user-id inspect [PATH]`
Symmetric sibling of `user-id create`: validates the file via the existing storage primitive, prints `user_id` + `user_words` (four-word-networking), supports `--json`, exits non-zero naming the file on failure. Smoke-tested end-to-end (text, JSON, missing-file). 4 unit tests.

### #90 — optional `domain` on `POST /agent/sign`
Signs `domain || 0x00 || payload` when `domain` present; echoes it in the response. Rejects NUL bytes, empty domain, and >1 KiB domains with 400. Omitted = byte-for-byte pre-#90 behavior (`scheme` unchanged — presence of `domain` in the response is the signal, per the issue's default option). CLI: `x0x agent sign --domain`. Also documented `/agent/sign` in `docs/api-reference.md` — it was missing from the reference entirely. 3 new daemon integration tests, including a negative test that the domain-separated signature does NOT verify over the raw payload.

### #85 — warn-only default logging
`default_log_level()` `info` → `warn` (and the invalid-value fallback likewise), so an x0xd started without `RUST_LOG`/config override logs warn/error only. `RUST_LOG=info` restores verbosity. README gains a Logging section. `/health` + `/diagnostics/*` visibility unaffected.

## Test plan
- [x] `cargo fmt --check`, `clippy --all-targets --all-features -D warnings`
- [x] `cargo doc` with `RUSTDOCFLAGS="-D warnings"`
- [x] Full workspace nextest 1680/1680
- [x] Daemon-backed `agent_sign` family with `--run-ignored` (7/7, includes 3 new #90 tests)
- [x] #101 regression test proven to fail pre-fix
- [x] `x0x user-id inspect` CLI smoke (text/JSON/error-exit)
- [ ] Testnet soak before merge (deploy + launch_soak windows) — running next

Closes #101. Closes #93. Closes #90. Closes #85.

<h3>Greptile Summary</h3>

This PR bundles four separate issue fixes: a cross-NAT DM \"black-hole\" caused by a lost watch channel send (#101), a new daemonless `user-id inspect` CLI command (#93), optional domain separation on `POST /agent/sign` (#90), and a `warn`-only default log level (#85).

- **#101 (critical correctness):** `start_dm_inbox` switched from `watch::Sender::send` to `send_replace`, which persists the capability upgrade in the channel even before any receiver subscribes — the root cause of peers advertising `gossip_inbox=false` forever across NAT. A regression test proving the pre-fix code fails is included.
- **#90 (domain separation):** `POST /agent/sign` accepts an optional `domain` string and signs `domain || 0x00 || payload`; the endpoint validates the domain (non-empty, NUL-free, ≤1 KiB) and echoes it back in the response. CLI and integration tests (including a negative replay test) are included.
- **#93 / #85:** New `user-id inspect` command and `warn`-default logging are clean, well-tested additions.

<h3>Confidence Score: 5/5</h3>

Safe to merge once the testnet soak window completes as noted in the PR checklist.

The core fix (send_replace) is a well-understood tokio::sync::watch API change with a regression test verified to fail on pre-fix code. Domain-separation logic is correct — NUL rejection, length capping, and the domain || 0x00 || payload construction are all sound, with integration tests verifying both happy path and negative replay. The remaining changes are self-contained and fully tested. No data-loss, auth, or correctness issues found.

No files require special attention; all changed paths are well-covered by the accompanying tests.

| Filename | Overview |
|----------|----------|
| src/lib.rs | Switches dm_capabilities_tx from send() to send_replace(), ensuring late subscribers see the capability upgrade; regression test included and proven to fail pre-fix. |
| src/bin/x0xd.rs | Adds domain-separation logic to agent_sign (with correct NUL/empty/length validation and domain || 0x00 || payload construction), changes default log level to warn, and updates the invalid-level fallback consistently. |
| src/cli/commands/user_id.rs | Adds inspect() function and InspectReport struct; four unit tests cover valid, missing, and malformed file paths. valid: bool is always true (failures return Err), documented and mirrors the ok: bool convention elsewhere. |
| src/cli/commands/identity.rs | Adds domain parameter to sign(), passes it as an optional JSON field; existing call sites updated to pass None; existing tests updated accordingly. |
| src/bin/x0x.rs | Wires up the new UserIdSub::Inspect variant and the --domain flag on AgentSub::Sign; both text and JSON output paths handled for inspect. |
| tests/daemon_api_integration.rs | Three new #[ignore] daemon integration tests covering domain-separated roundtrip, no-domain-field-in-response, and NUL-byte rejection; roundtrip test also verifies the signature does NOT verify over raw payload. |
| docs/api-reference.md | Documents POST /agent/sign endpoint and domain field that were previously missing from the API reference. |
| README.md | Adds a Logging section explaining the warn-only default and how to opt in to verbose output via RUST_LOG. |

</details>

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant xd as x0xd startup
    participant inbox as start_dm_inbox
    participant watch as watch::Sender<DmCapabilities>
    participant advert as capability_advert_service

    xd->>inbox: start_dm_inbox(kem_keypair)
    inbox->>watch: "send_replace(upgraded caps)<br/>gossip_inbox=true, kem_key=..."
    Note over watch: value stored in channel<br/>even with no subscribers yet
    xd->>advert: start_capability_advert_service (subscribes late)
    advert->>watch: subscribe() → borrow()
    watch-->>advert: "upgraded caps (gossip_inbox=true)"
    advert-->>advert: "advertise gossip_inbox=true to peers"

    Note over xd,advert: Pre-fix: send() dropped value<br/>advert saw gossip_inbox=false → NAT black-hole
```

<sub>Reviews (1): Last reviewed commit: ["fix(dm): store capability upgrade with s..."](https://github.com/saorsa-labs/x0x/commit/0b0a1a7bc2b5354873c339258c3a9147c8c00545) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36654750)</sub>

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,6 +314,26 @@ Each instance gets its own identity, port, and data directory.
 
 ---
 
+## Logging
+
+`x0xd` is quiet by default: when neither `RUST_LOG` nor the config
+`log_level` is set, only `warn` and `error` lines are emitted. This is a
+privacy default — verbose levels include peer and topic activity that an
+operator may not want written to logs.
+
+Opt in to verbose logging explicitly:
+
+```bash
+RUST_LOG=info x0xd            # standard verbosity
+RUST_LOG=debug x0xd           # full debugging
+RUST_LOG=ant_quic=debug x0xd  # per-module filters work too
+```
+
+Operator visibility via `GET /health` and `GET /diagnostics/*` is
+independent of the log level.
+
+---
+
 ## GUI
 
 x0x includes a built-in web interface. No download, no install — it's embedded in the binary.
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -68,6 +68,7 @@ curl http://127.0.0.1:12700/status
 | GET | `/agent/user-id` | `x0x agent user-id` | Current user ID if configured |
 | GET | `/agent/card` | `x0x agent card` | Generate a shareable identity card |
 | POST | `/agent/card/import` | `x0x agent import` | Import a card into contacts |
+| POST | `/agent/sign` | `x0x agent sign` | Detached ML-DSA-65 signature over caller-supplied bytes |
 
 ### Announce request body
 
@@ -85,6 +86,26 @@ Notes:
 ### Agent card query params
 
 `GET /agent/card?display_name=Alice&include_groups=true`
+
+### Sign request body
+
+```json
+{
+  "payload_b64": "<base64 bytes to sign>",
+  "domain": "my-protocol.v1.register"
+}
+```
+
+Notes:
+- `payload_b64` is decoded and signed verbatim (max 256 KiB). Callers
+  canonicalize structured payloads themselves.
+- `domain` is optional. When present, the daemon signs
+  `domain || 0x00 || payload` and echoes `domain` in the response, so a
+  signature for one protocol context cannot be replayed in another.
+  Domains must be non-empty, NUL-free, and at most 1024 bytes.
+  `x0x.<purpose>.<version>` is the conventional shape for x0x protocols.
+- Response: `ok`, `agent_id` (hex), `public_key_b64`, `signature_b64`,
+  `algorithm` (`x0x.agent-sign.v1.ml-dsa-65`), and `domain` when supplied.
 
 ## Network
 
--- a/src/bin/x0x.rs
+++ b/src/bin/x0x.rs
@@ -331,6 +331,9 @@ enum AgentSub {
         /// Base64-encoded bytes to sign.
         #[arg(long)]
         payload_b64: Option<String>,
+        /// Optional domain-separation string; signs `domain || 0x00 || payload`.
+        #[arg(long)]
+        domain: Option<String>,
     },
 }
 
@@ -340,6 +343,12 @@ enum UserIdSub {
     /// Overwrites any existing file at the target path without prompting.
     Create {
         /// Output path. Existing file at this path is overwritten.
+        path: Option<PathBuf>,
+    },
+    /// Read and validate a user identity file (no daemon needed).
+    /// Defaults to ~/.x0x/user.key.
+    Inspect {
+        /// Path of the key file to inspect.
         path: Option<PathBuf>,
     },
 }
@@ -1113,6 +1122,22 @@ async fn run(
                 }
                 return Ok(());
             }
+            UserIdSub::Inspect { path } => {
+                let report = commands::user_id::inspect(path.clone()).await?;
+                match format {
+                    OutputFormat::Json => {
+                        x0x::cli::print_value(format, &serde_json::to_value(&report)?)
+                    }
+                    OutputFormat::Text => {
+                        println!("User identity at {}:", report.path);
+                        println!("user_id:    {}", report.user_id);
+                        if let Some(words) = &report.user_words {
+                            println!("user_words: {words}");
+                        }
+                    }
+                }
+                return Ok(());
+            }
         },
         _ => {}
     }
@@ -1173,8 +1198,18 @@ async fn run(
             Some(AgentSub::Import { card, trust }) => {
                 commands::identity::import_card(&client, &card, Some(trust.as_str())).await
             }
-            Some(AgentSub::Sign { file, payload_b64 }) => {
-                commands::identity::sign(&client, file.as_deref(), payload_b64.as_deref()).await
+            Some(AgentSub::Sign {
+                file,
+                payload_b64,
+                domain,
+            }) => {
+                commands::identity::sign(
+                    &client,
+                    file.as_deref(),
+                    payload_b64.as_deref(),
+                    domain.as_deref(),
+                )
+                .await
             }
         },
         Commands::Announce {
@@ -1649,6 +1684,7 @@ x0x (v{VERSION})
 |   |   +-- card           Generate shareable identity card
 |   |   +-- import         Import an agent card to contacts
 |   +-- user-id create     Create user identity keypair
+|   +-- user-id inspect    Validate a user identity file (daemonless)
 |   +-- announce           Announce identity to network
 |
 +-- Network
--- a/src/bin/x0xd.rs
+++ b/src/bin/x0xd.rs
@@ -265,7 +265,10 @@ fn shared_cache_dir() -> PathBuf {
 }
 
 fn default_log_level() -> String {
-    "info".to_string()
+    // Privacy by default for operators outside our fleet (issue #85): without
+    // an explicit RUST_LOG or config log_level override, x0xd logs warn/error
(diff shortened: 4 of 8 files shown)


dirvine:
## Testnet soak evidence (Batch 1)

Deployed to all 6 testnet nodes (24/24 deploy checks, 12–13 peers/node), then 4 × 30-min `launch_soak.py` windows against the broad-launch gate (anchor nyc):

| window | Phase A (DM delivery) | drop_full Δ | verdict |
|---|---|---|---|
| 1 | **30/30** | 223 | NO-GO |
| 2 | **30/30** | 10,887 | NO-GO |
| 3 | **30/30** | 6,778 | NO-GO |
| 4 | **30/30** | 6,228 | NO-GO |

**The NO-GO is environmental, not this PR.** Verified directly:

- The co-located **prod v0.21.4 daemons (zero Batch-1 code, untouched by this deploy)** show the same drop regime: nyc prod has ~300k cumulative admission/pump drops since last night's auto-upgrade boot and is actively dropping at ~2k counts/min, with `dispatcher.membership.timed_out=13,719`.
- **Same-host A/B over the same 2 minutes: prod dropped 6,238 vs this build's 2,785** — the Batch-1 build drops at *less than half* prod's rate under identical conditions (plausibly because #101 moves DMs onto the gossip-inbox path instead of raw-QUIC retries).
- Functional delivery is unimpaired: Phase A 30/30 in every window; prod journal shows zero warnings; peers stable.

The fleet-wide drop regime (admission drops to dead/suspect peers from a churning remote-peer population) needs its own investigation and likely a drop_full gate recalibration — tracked separately from this PR. Soak artifacts: `proofs/launch-readiness-soak-20260610T080701Z/`.

Merging on: CI 15/15 green, full workspace nextest 1680/1680, regression test proven fail-pre-fix, Phase A 30/30 × 4, A/B non-regression vs prod.

PR #103, opened 2026-06-10
Batch 2: KvStore first-join bootstrap (#96), local: topics (#89), seeded user keygen (#95), logging redaction module (#83 groundwork)

### LINKED ISSUE #89: Proposal: `local:` topic prefix for non-gossipped pub/sub
# Proposal: `local:` topic prefix for non-gossipped pub/sub

## Summary

Add a `local:` topic prefix to `x0xd`'s pub/sub. Topics with this prefix are NOT relayed to remote mesh peers — they are delivered only to subscribers attached to the same `x0xd` instance. All other primitives (`/publish`, `/subscribe`, `/events`, bearer-token auth, verified/trust annotations) work unchanged.

## Motivation

`x0xd` is increasingly used as a local-IPC substrate by multi-process applications running on a single machine — the daemon + apps pattern documented in `docs/local-apps.md`. These apps want the ergonomic of pub/sub for fan-out and event streaming between sibling processes, but they don't want those events broadcast to the global mesh.

Today there is no clean way to do this with the existing primitive. Three workarounds are observable, each with problems:

1. **Roll your own loopback HTTP server.** The app implements its own server on `127.0.0.1`, reimplementing subscribe/publish/auth. Duplicates work `x0xd` already does well, and the surface is easy to get wrong (CSRF, token rotation, etc.).
2. **Publish to globally-named topics anyway.** Either accepting the leak, or hoping no one subscribes. Privacy footgun.
3. **Use `/direct/send` for point-to-point.** Works for 1-to-1 but doesn't fan out, and doesn't carry the verified/trust annotations gossip does.

The clean solution is a routing-mode opt-in on the existing primitive.

## Proposal

A topic name that starts with `local:` is recognized by `x0xd` as a routing-mode hint. Such topics are:

- Delivered to all locally-subscribed callers on this `x0xd` instance.
- NOT forwarded into the PlumTree EAGER set.
- NOT included in `IHAVE` digests sent to remote peers.
- NOT relayed when an inbound IHAVE references such a topic (defensive — even if a misbehaving peer tries to gossip a `local:` topic into the mesh, drop it).

Everything else stays identical: same `/publish` / `/subscribe` / `/events` endpoints, same bearer-token gate, same SSE event shape, same `verified` / `trust_level` annotations (for `local:` topics, the trust signal is just "the bearer-token holder," same as any local API caller).

## API examples

Publish:

```http
POST /publish
Authorization: Bearer <token>
Content-Type: application/json

{
  "topic": "local:my-app/events",
  "payload": "eyJ0eXBlIjoicGluZyJ9"
}
```

Subscribe:

```http
POST /subscribe
Authorization: Bearer <token>
Content-Type: application/json

{
  "topic": "local:my-app/events"
}
```

Receive (`GET /events` SSE stream):

```json
{
  "type": "message",
  "data": {
    "subscription_id": "...",
    "topic": "local:my-app/events",
    "payload": "...",
    "sender": "...",
    "verified": true,
    "trust_level": "known"
  }
}
```

### Concrete use case — sibling-process IPC

```text
Daemon subscribes to: local:my-app/commands
App-A publishes to:    local:my-app/commands  →  daemon receives, mesh does not.
App-B subscribes to:   local:my-app/status
Daemon publishes to:   local:my-app/status    →  App-B receives, mesh does not.
```

Replaces the loopback-HTTP-server pattern in `docs/local-apps.md` for apps that want pub/sub semantics on local IPC.

## Implementation sketch

Single classifier branch in the dispatch path. Approximate location: `src/gossip/plumtree.rs` (the publish-local and inbound-relay paths).

```rust
fn is_local_topic(topic: &str) -> bool {
    topic.starts_with("local:")
}

// In publish path:
if is_local_topic(&topic) {
    deliver_to_local_subscribers_only(topic, payload);
    return; // skip EAGER fan-out
}

// In inbound-relay path:
if is_local_topic(&topic) {
    // a remote peer should NEVER send us a local: topic; drop + log
    return;
}
```

The admission-control classifier already in `register_x0x_topic_priorities()` stays unchanged — `local:` topics are short-circuited before they reach it.

Subscription bookkeeping likely doesn't need to change; subscribers to `local:` topics are just subscribers, the relay layer is where the routing differs.

Estimated diff: ~20–40 lines in PubSubManager + 4 unit tests:
- local publish delivers locally
- local publish doesn't enter EAGER
- inbound `local:` from a remote peer is dropped
- SSE event shape is unchanged

## Non-goals

- **Not a new security boundary.** The bearer token still gates the local API; `local:` doesn't add per-app permissioning. Apps that need stronger isolation between sibling processes layer their own crypto on top (sealed envelopes, etc.). The proposal restricts BROADCAST scope, not LOCAL access.
- **Not a new auth model.** No new tokens, no scoped permissions.
- **Not a substitute for direct messaging.** `local:` is for fan-out to multiple local subscribers. `/direct/send` is still the right primitive for point-to-point.

## Compatibility

- **Fully additive.** Existing topics (no prefix) keep their current global-relay behavior.
- **No wire format change.** Same JSON shapes.
- **No new endpoints.**
- **Existing consumers unaffected.** A consumer who doesn't use `local:` sees no change.

## Alternatives considered

- **`delivery: "local"` request field instead of a prefix.** Equivalent expressiveness. Topic-prefix is preferred because (a) the topic name is visible in logs / diagnostics / SSE events, so the routing mode is self-documenting, and (b) a subscriber accidentally subscribing to a publisher's local topic FROM A REMOTE PEER would silently never receive — with a prefix, the impossibility is visible in the name.
- **Per-publish `relay: false` boolean.** Same trade-offs as the field option above.
- **Reuse `chat_topic` of a self-owned `MlsEncrypted` group of one member.** Doesn't compose — groups expect >1 member; secret rotation logic doesn't apply; UX is wrong for "local IPC."

## Prior art

- `docs/local-apps.md` documents the local-IPC pattern but doesn't address how to use pub/sub safely from it.
- The admission-control classifier (`register_x0x_topic_priorities` in `src/gossip/runtime.rs`) is the natural sibling for the routing-mode classifier.
- PR #88 (encrypted KvStore) addresses shared-state-in-a-group; orthogonal to this proposal — composes if a `local:`-scoped encrypted store is ever wanted.

## Asks

- Maintainer LGTM on the prefix syntax (`local:` vs `local/`).
- Confirmation that the classifier-branch location in PlumtreePubSub is the right insertion point.
- Open to alternative names if `local:` collides with existing conventions.

dirvine:
## Summary

Second batch of the open-issue cleanup. Stacked on #102 (Batch 1) — contains its commits; merge #102 first and this rebases to just the Batch-2 commit.

### #96 — KvStore first-time late join never bootstraps pre-existing keys
Root cause: a first-time joiner only ever received deltas published after it subscribed. The gossip message cache replays ~60s, and its lazy pruning (runs on insert) means *busy* topics lose older deltas entirely — matching the issue's scenario 3 exactly (future writes arrived in ~1s, the historical key never did).

Fix: a `<topic>/state-sync` side channel. Empty-store joiners publish `StateRequest`s on a 1/5/15/30s schedule; any holder with state responds by republishing `full_delta()` as a normal CRDT delta on the main topic (idempotent merge, multi-holder safe). Backward compatible: the main-topic wire format is untouched and pre-fix nodes simply never subscribe to the side channel.

**Proof**: new cold-join daemon test (alice writes while bob's daemon does not exist; 70s cache-expiry wait; busy-topic writes force the prune; bob boots, connects, joins) — **fails against the pre-fix binary** (`X0XD_TEST_BINARY` override: bob receives the fresh keys, never the historical one — the issue's exact signature) and **passes post-fix** in ~20s. A second test guards the join-before-write live path.

### #89 — `local:` topics (same-daemon pub/sub IPC)
Topics prefixed `local:` are handled in `PubSubManager` with a local subscriber registry and never touch PlumTree — no EAGER set, no IHAVE digests, and inbound gossip cannot deliver to local subscribers because no PlumTree subscription ever exists. `/publish`, `/subscribe`, WS subscribe, SSE, bearer auth all unchanged. Documented in the API reference.

**Tests**: daemon-pair WS tests prove same-daemon delivery, two-subscriber fan-out, and that a *connected remote peer subscribed to the identical topic name receives nothing*.

### #95 — `x0x user-id create --from-seed <hex>`
Deterministic UserKeypair via FIPS 204 seeded KeyGen — `fips204::ml_dsa_65::KG::keygen_from_seed` (the ξ input; the crate was already a direct dependency, so no upstream work was needed). Same 32-byte seed → same keypair on any machine; BIP-39 mnemonic encoding intentionally stays in consumer apps per the issue's separation-of-concerns table. Validation: hex/length checks (no file written on bad seed), determinism + distinct-seed + sign/verify tests.

### #83 — groundwork only
`src/logging.rs`: salted-BLAKE3 identifier wrappers (`LogAgentId`/`LogMachineId`/`LogUserId`/`LogHexId`, re-exported upstream `LogPeerId`/`LogTopicId`), per-process memory-only salt, Debug mirrors Display so `?` can't leak. The ~48-site warn!/error! call-site migration ships as its own follow-up PR for reviewability.

### Test-harness hardening
- `solo()` / `join_peer()` staggered-start helpers (state-before-the-joiner-exists scenarios)
- `X0XD_TEST_BINARY` runtime override (lets a regression test run against a pinned pre-fix daemon)
- daemon health deadline 30s→90s (debug builds + hard-coded internet bootstrap rounds made 30s a first-daemon flakiness knife-edge)
- new test files wired into `integration.yml` (single-threaded)

## Test plan
- [x] fmt / clippy `-D warnings` / `cargo doc -D warnings` clean
- [x] Full workspace nextest **1687/1687**
- [x] 4/4 daemon-backed tests (serial)
- [x] #96 test proven to fail on pre-fix binary
- [ ] Testnet soak after #102 merges (note: fleet-wide degradation today affects prod 0.21.4 equally — A/B drop-rate comparison on the same host shows the new build at *less than half* prod's drop rate)

Closes #96. Closes #89. Closes #95.

<h3>Greptile Summary</h3>

This batch closes issues #96 (KvStore cold-join bootstrap via a state-sync side channel), #89 (`local:` topics for same-daemon pub/sub IPC), and #95 (deterministic `UserKeypair` from a 32-byte seed), plus groundwork for #83 (salted-BLAKE3 log-identifier redaction) and a domain-separation field on the agent-sign endpoint.

- **#96 / `src/kv/sync.rs`**: Empty-store joiners publish `StateRequest`s on `<topic>/state-sync` at 1/5/15/30 s; any holder with non-empty state responds by republishing `full_delta()` as a regular CRDT delta on the main topic. Backward-compatible wire format; the 70 s cache-expiry integration test and pre-fix binary regression proof are solid, but the access-control interaction between `merge_delta`'s `writer` check and non-owner responders is a gap for `Signed`/`Allowlisted` stores (see inline comment).
- **#89 / `src/gossip/pubsub.rs`**: `local:` topics are routed through a per-`PubSubManager` `HashMap<String, Vec<Sender>>`, bypassing PlumTree entirely; fan-out, lazy dead-sender cleanup, and isolation from remote peers are all verified by new daemon-pair WS tests.
- **#95 / `src/identity.rs` + `src/cli/commands/user_id.rs`**: `UserKeypair::from_seed` wraps `fips204::ml_dsa_65::KG::keygen_from_seed`; determinism, distinct-seed, and sign/verify properties are unit-tested; the CLI validates hex/length before writing any file.

<h3>Confidence Score: 3/5</h3>

The local-topic routing, seeded keygen, domain-separation signing, and test-harness changes are solid. The state-sync bootstrap works for the tested 2-node owner-is-online scenario but silently breaks for Signed and Allowlisted stores when only non-owner replicas can respond.

The state-sync responder publishes full_delta() signed with its own agent keypair, but merge_delta on the receiving side runs the same is_authorized() check used for live writes. For Signed stores (the default policy), only the owner's agent_id passes that check — responses from every other replica are silently dropped. The integration test only exercises Alice-as-owner responding to Bob, so the failure mode does not appear in CI.

src/kv/sync.rs — how merge_delta writer-authorization interacts with state-sync responses from non-owner holders. src/logging.rs — init_log_salt_for_tests needs a cfg(test) guard.

| Filename | Overview |
|----------|----------|
| src/kv/sync.rs | Adds state-sync side channel (#96): responder and bootstrap-requester tasks are correct for the 2-node owner-is-online case but silently fail for non-owner responders on Signed/Allowlisted stores, and the requester task is not cancelled on stop(). |
| src/gossip/pubsub.rs | Adds local: topic routing with a centralized sender-Vec registry; fan-out, lazy dead-sender cleanup via retain, and unsubscribe semantics look correct and consistent with the existing PlumTree subscriber pattern. |
| src/logging.rs | New privacy-preserving log-identifier module: per-process BLAKE3-keyed salt, correct Debug=Display guard, well-tested — init_log_salt_for_tests needs a cfg(test) guard to prevent production exposure. |
| src/identity.rs | Adds UserKeypair::from_seed using fips204 seeded KeyGen; deterministic, sign/verify tested, error path is clean. |
| src/cli/commands/user_id.rs | Adds --from-seed hex validation (length + decode), inspect subcommand, and determinism/rejection/malformed-file tests — all paths are well-covered and no file is written on invalid input. |
| src/bin/x0xd.rs | Domain-separation signing (#90): empty/NUL/length checks are correct, canonical bytes construction is unambiguous, domain echoed in response, and send_replace fix for #101 prevents the DM capability race. |
| tests/harness/src/cluster.rs | Adds solo()/join_peer() staggered-start helpers, X0XD_TEST_BINARY override for pre-fix regression proofs, and bumps the health deadline to 90s. |
| tests/kv_first_join_bootstrap.rs | Regression test for issue #96 using solo/join_peer and a 70s cache-expiry wait; proves the failure scenario. Only covers the 2-node, owner-is-online path. |
| tests/local_topics.rs | Daemon-backed tests prove same-daemon delivery, two-subscriber fan-out, and non-leakage to a connected remote peer — complete coverage of the local: topic contract. |
| src/lib.rs | Adds local logging module export, fixes DM capability upgrade to send_replace (#101), and passes local_peer_id to KvStoreSync::new. |

</details>

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Alice as Alice (owner daemon)
    participant Mesh as PlumTree Mesh
    participant Bob as Bob (late-joiner daemon)

    Alice->>Mesh: PUT /stores/topic/key (publish KvStoreDelta)
    Note over Mesh: >60s pass, cache pruned

    Bob->>Mesh: subscribe(topic) + subscribe(topic/state-sync)
    Note over Bob: store is_empty() == true, spawn requester

    loop STATE_REQUEST_RETRY_SECS 1,5,15,30
        Bob->>Mesh: publish(topic/state-sync, StateRequest)
        Mesh->>Alice: deliver StateRequest
        Alice-->>Mesh: publish(topic, full_delta)
        Mesh-->>Bob: deliver full_delta
        Bob->>Bob: "merge_delta writer=alice_agent_id"
        Note over Bob: is_authorized(alice) true for Signed store
    end

    Note over Bob: local:topic messages stay on-daemon only
    Bob->>Bob: subscribe local:events to local_topics map
    Bob->>Bob: publish local:events via retain try_send
    Note over Bob: PlumTree never sees local: frames
```

`````markdown

---

### Issue 1 of 3
src/kv/sync.rs:157-165
**State-sync bootstrap silently rejected for non-owner responders**

When a holder republishes `full_delta()` as a state-sync response it signs the gossip message with its own agent keypair. On the joiner's side, `merge_delta` uses `msg.sender` (the responder's `AgentId`) as the `writer` parameter. For `AccessPolicy::Signed` stores (the default), `is_authorized` only returns true for the store owner — so a full-delta republished by any non-owner replica is silently rejected (`return Ok(())` in `KvStore::merge_delta`). Bootstrap therefore works only if the store owner happens to be among the responding holders. In a scenario where the owner is offline but replicas are available, the joiner never bootstraps, which is precisely the failure mode issue #96 aimed to fix.

### Issue 2 of 3
src/kv/sync.rs:180-200
**Requester task outlives `stop()` and can double-fire if `start()` is re-called**

The requester task captures only `Arc<PubSubManager>` and the sync topic string, so it is not cancelled when `stop()` calls `unsubscribe`. The task continues sleeping and publishing for up to 51 s (1+5+15+30) after the store is torn down. If the same `KvStoreSync` (or a replacement for the same topic) calls `start()` again before the old task finishes, two requester tasks run concurrently, doubling request volume during the overlap window.

### Issue 3 of 3
src/logging.rs:57-60
**Test helper exposed in production builds**

`init_log_salt_for_tests` is `pub` without a `#[cfg(any(test, feature = "testing"))]` guard. Any code that imports `x0x::logging` and runs in the same process can call it before the first log emission, fixing the `OnceLock` salt to a known value. After that, every identifier hashes to a predictable token, defeating the per-session privacy guarantee the module provides. Consider gating this behind `#[cfg(test)]` or a dedicated `testing` feature flag.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(kv,pubsub,cli): KvStore first-join ..."](https://github.com/saorsa-labs/x0x/commit/7f07733627efbeaffab0b0c1cc931b9d622dbf70) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36668123)</sub>

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -85,6 +85,10 @@ jobs:
         run: cargo nextest run --all-features --test gui_smoke
       - name: Run WebSocket tests
         run: cargo nextest run --all-features --test ws_integration -- --ignored
+      - name: Run KvStore first-join bootstrap tests (issue 96)
+        run: cargo nextest run --all-features --test kv_first_join_bootstrap --test-threads 1 -- --ignored
+      - name: Run local-topic routing tests (issue 89)
+        run: cargo nextest run --all-features --test local_topics --test-threads 1 -- --ignored
 
   # ── Soak test (weekly, ~90min) ────────────────────────────────────────
   soak:
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -134,6 +134,16 @@ Notes:
 }
 ```
 
+### `local:` topics (same-daemon IPC)
+
+Topics whose name starts with `local:` (e.g. `local:my-app/events`) are
+never gossipped: messages are delivered only to subscribers attached to
+the same `x0xd` instance — they never enter the PlumTree EAGER set or
+IHAVE digests. All primitives work unchanged (`/publish`, `/subscribe`,
+`/events`, WebSocket subscribe, bearer-token auth). Use them as a local
+pub/sub substrate for multi-process applications sharing one daemon,
+without leaking events to the mesh.
+
 ## Discovery
 
 | Method | Endpoint | CLI | Purpose |
--- a/src/bin/x0x-user-keygen.rs
+++ b/src/bin/x0x-user-keygen.rs
@@ -15,7 +15,7 @@ async fn main() -> Result<()> {
         .map(PathBuf::from)
         .context("usage: x0x-user-keygen <output-path>")?;
 
-    let resolved = x0x::cli::commands::user_id::create(Some(output)).await?;
+    let resolved = x0x::cli::commands::user_id::create(Some(output), None).await?;
     println!("{}", resolved.display());
     Ok(())
 }
--- a/src/bin/x0x.rs
+++ b/src/bin/x0x.rs
@@ -344,6 +344,10 @@ enum UserIdSub {
     Create {
         /// Output path. Existing file at this path is overwritten.
         path: Option<PathBuf>,
+        /// Derive the keypair deterministically from a 32-byte hex seed
+        /// (64 hex chars) via FIPS 204 seeded KeyGen. Same seed, same keypair.
+        #[arg(long, value_name = "HEX")]
+        from_seed: Option<String>,
     },
     /// Read and validate a user identity file (no daemon needed).
     /// Defaults to ~/.x0x/user.key.
@@ -1109,8 +1113,9 @@ async fn run(
             };
         }
         Commands::UserId { sub } => match sub {
-            UserIdSub::Create { path } => {
-                let resolved = commands::user_id::create(path.clone()).await?;
+            UserIdSub::Create { path, from_seed } => {
+                let resolved =
+                    commands::user_id::create(path.clone(), from_seed.as_deref()).await?;
                 match format {
                     OutputFormat::Json => x0x::cli::print_value(
                         format,
--- a/src/cli/commands/user_id.rs
+++ b/src/cli/commands/user_id.rs
@@ -13,14 +13,32 @@ pub fn default_user_key_path() -> Result<PathBuf> {
 }
 
 /// `x0x user-id create [PATH]` — create a user identity keypair on disk.
-pub async fn create(output: Option<PathBuf>) -> Result<PathBuf> {
+///
+/// With `from_seed_hex` (issue #95), the keypair is derived
+/// deterministically from the 32-byte seed via FIPS 204 seeded KeyGen:
+/// same seed → same keypair on any machine. Without it, random keygen.
+pub async fn create(output: Option<PathBuf>, from_seed_hex: Option<&str>) -> Result<PathBuf> {
     let path = match output {
         Some(path) => path,
         None => default_user_key_path()?,
     };
 
-    let keypair = UserKeypair::generate()
-        .map_err(|e| anyhow::anyhow!("failed to generate user keypair: {e}"))?;
+    let keypair = match from_seed_hex {
+        Some(hex_seed) => {
+            let bytes = hex::decode(hex_seed.trim())
+                .context("--from-seed must be hex (64 hex chars encoding 32 bytes)")?;
+            let seed: [u8; 32] = bytes.as_slice().try_into().map_err(|_| {
+                anyhow::anyhow!(
+                    "--from-seed must encode exactly 32 bytes (64 hex chars), got {} bytes",
+                    bytes.len()
+                )
+            })?;
+            UserKeypair::from_seed(&seed)
+                .map_err(|e| anyhow::anyhow!("failed to derive user keypair from seed: {e}"))?
+        }
+        None => UserKeypair::generate()
+            .map_err(|e| anyhow::anyhow!("failed to generate user keypair: {e}"))?,
+    };
     save_user_keypair_to(&keypair, &path)
         .await
         .with_context(|| format!("failed to write user keypair to {}", path.display()))?;
@@ -80,7 +98,7 @@ mod tests {
         let temp_dir = tempfile::tempdir()?;
         let path = temp_dir.path().join("user.key");
 
-        let resolved = create(Some(path.clone())).await?;
+        let resolved = create(Some(path.clone()), None).await?;
 
         ensure!(
             resolved == path,
@@ -101,7 +119,7 @@ mod tests {
     async fn inspect_reports_user_id_and_words_for_valid_file() -> Result<()> {
         let temp_dir = tempfile::tempdir()?;
         let path = temp_dir.path().join("user.key");
-        create(Some(path.clone())).await?;
+        create(Some(path.clone()), None).await?;
         let expected = hex::encode(load_user_keypair_from(&path).await?.user_id().as_bytes());
 
         let report = inspect(Some(path.clone())).await?;
@@ -120,6 +138,52 @@ mod tests {
             "report must echo the inspected path"
         );
         Ok(())
+    }
+
+    #[tokio::test]
+    async fn create_from_seed_is_deterministic_across_machines() -> Result<()> {
+        // Issue #95: same seed must yield the same UserKeypair every time —
+        // this is the contract mnemonic-based identity portability rests on.
+        let temp_dir = tempfile::tempdir()?;
+        let seed_hex = "0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20";
+
+        let path_a = temp_dir.path().join("a.key");
+        let path_b = temp_dir.path().join("b.key");
+        create(Some(path_a.clone()), Some(seed_hex)).await?;
+        create(Some(path_b.clone()), Some(seed_hex)).await?;
+
+        let id_a = load_user_keypair_from(&path_a).await?.user_id();
+        let id_b = load_user_keypair_from(&path_b).await?.user_id();
+        ensure!(id_a == id_b, "same seed must derive the same user_id");
+
+        let path_c = temp_dir.path().join("c.key");
+        let other_seed = "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff";
+        create(Some(path_c.clone()), Some(other_seed)).await?;
+        let id_c = load_user_keypair_from(&path_c).await?.user_id();
+        ensure!(
+            id_a != id_c,
+            "different seeds must derive different user_ids"
+        );
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn create_from_seed_rejects_bad_seed_encoding() {
+        let temp_dir = tempfile::tempdir().expect("tmpdir");
+        let path = temp_dir.path().join("user.key");
+
+        assert!(
+            create(Some(path.clone()), Some("not-hex")).await.is_err(),
+            "non-hex seed must be rejected"
+        );
+        assert!(
+            create(Some(path.clone()), Some("aabb")).await.is_err(),
+            "short seed must be rejected"
+        );
+        assert!(
+            !path.exists(),
+            "no key file may be written when the seed is invalid"
+        );
     }
 
     #[tokio::test]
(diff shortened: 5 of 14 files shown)


dirvine:
## Testnet soak evidence (Batch 2)

Deployed to all 6 testnet nodes (24/24 deploy checks), then 2 × 30-min `launch_soak.py` windows (broad-launch gate, anchor nyc):

| window | Phase A (DM delivery) | drop_full Δ | verdict |
|---|---|---|---|
| 1 | **28/28 (100%)** | 840 | NO-GO |
| 2 | **29/29 (100%)** | 1,952 | NO-GO |

Same environmental NO-GO as #102 (see its soak comment for the full attribution): the fleet-wide drop regime affects prod v0.21.4 equally. **Same-host A/B over the same 2 minutes: prod 7,093 vs this build 7,096** — statistically identical, i.e. zero regression from the Batch-2 changes. Batch-2's drop_full deltas (840/1,952) are actually the lowest of any window measured today.

The new daemon-backed tests for #96 (cold first-join bootstrap, proven fail-pre-fix) and #89 (local: no-leak) also ran green in this PR's Multi-Agent Integration CI job.

Merging on: CI fully green post-rebase, workspace nextest 1687/1687, Phase A 100% × 2, A/B parity with prod. Soak artifacts: `proofs/launch-readiness-soak-20260610T095605Z/`.

PR #1, opened 2026-02-09
feat: optional CRDT persistence with checkpoint/recovery and binding parity

JimCollinson:
## Summary

This PR adds an optional local persistence layer for protocol-managed CRDT state in x0x, with strict/degraded behaviour, deterministic recovery semantics, and Node/Python API parity.

Persistence exists so agents can survive restart and rejoin the gossip network from where they left off, rather than requiring full state reconciliation from peers. It is deliberately optional and degraded-by-default because the agent's primary job is network participation - a disk failure should not stop an agent from cooperating. Persistence is a local node preference; it does not affect wire protocol or peer interoperability.

This is a library-layer delivery. The persistence subsystem components are complete and tested, but automatic Agent lifecycle wiring (startup recovery and checkpoint orchestration at the top-level agent flow) will follow in a separate PR.

## Scope Delivered

- Optional persistence contracts and runtime policy surface, preserving existing in-memory behaviour when persistence is disabled.
- File-snapshot backend with atomic writes, manifest-based initialisation semantics, migration compatibility checks, retention/budget controls, and deterministic recovery scan behaviour.
- Checkpoint and recovery primitives: startup load/rejoin, explicit checkpoint requests, graceful shutdown checkpoint handling, and health/observability surfaces.
- Node/Python binding parity for persistence config, observability payloads (`health`, checkpoint frequency, bounds), and bounded runtime control requests.
- Hardening and hygiene: path traversal protection, recovery iteration correctness, module-surface cleanup, backup artifact hygiene.

Implementation defaults are documented in `docs/persistence.md` and architecture rationale in `docs/decisions.md`.

## Security Boundary

Persisted protocol state is plaintext by design. Security boundary is filesystem permissions, OS-level controls, and optional full-disk encryption.

## Architectural Decisions (in-repo)

This PR includes decision rationale in `docs/decisions.md` so contributors can review design intent alongside the code.

## How to Review

Primary command:
- `cargo test --workspace`

Suggested focal files:

- `src/crdt/persistence/backends/file_backend.rs` - backend, recovery iteration, path safety
- `src/crdt/persistence/orchestrator.rs` - startup/shutdown orchestration
- `src/runtime/agent_api.rs` - agent-facing checkpoint API and health transitions
- `src/runtime/policy_bounds.rs` - host policy envelope enforcement
- `bindings/nodejs/src/{config,health,runtime_controls}.rs` - Node binding parity
- `bindings/python/src/{config,health,runtime_controls}.rs` - Python binding parity
[...]

PR #121, opened 2026-06-25
ADR-0016 Phase 1 — role-based group authority: retire `Owner` authority (flat Admin/Member)

JimCollinson:
> Implements **Phase 1 of [#107](https://github.com/saorsa-labs/x0x/issues/107) / [ADR-0016](https://github.com/saorsa-labs/x0x/blob/main/docs/adr/0016-role-based-group-authority-flat-admin.md)** — the authority-alignment phase set out on #107 (parts **(a)** and **(c)**). Authority for named groups is decided by the signer's **role on the committed roster**; the `Owner` *role authority* and every creator-identity gate are retired in favour of a flat **Admin / Member** model where **Admin is root for the group**. The `Owner` enum variant, its serialization and rendering remain for legacy compatibility.
 
## Summary
 
- **What & why.** Per #107 / ADR-0016, authority is decided by committed-roster role in `validate_apply` — never by `GroupInfo.creator`. `Owner`/`OwnerOnly`/`require_owner` are retired; any Admin may promote, demote, remove, invite, change policy, or delete the group (*ban authority is Admin-gated, but non-creator-admin TreeKEM ban remains blocked by the existing KeyPackage/424 boundary until Phase 2 — see Deferred*). A **last-admin invariant** prevents a live group reaching zero admins. `DELETE /groups/:id` becomes a pure **self-leave**; ending the group is the any-admin **delete** action over the existing `POST /groups/:id/state/withdraw`, which marks a **keyless tombstone** and wipes all key material.
- **Base.** Prepared on **v0.26.0**; confirmed a **clean fast-forward** on current `main` (`a6fce96`, no upstream movement) — worth a final re-check at open time.
- **No wire/format change** — behaviour + surfaces only; legacy chains verify byte-for-byte (details below).
- **On the diff size.** ~80% of the `+10,174` is tests; the shipped code grows ~1,900 lines — all of it the ADR-0012 delete-hardening subsystem — while the authority change itself is a net deletion. Full breakdown below.
## Validation
 
- [x] `cargo fmt --all -- --check` — green in CI
- [x] `cargo clippy --all-targets --all-features -- -D warnings` — green in CI
- [x] `cargo nextest run --all-features --workspace` — green in CI (full Test Suite + Multi-Agent Integration + Property Tests + API Coverage Guard + Security Audit, across linux x64/arm64 gnu+musl and macOS x64/arm64). *Integration & Soak Tests / Soak Test is skipped by the workflow — the upstream endurance gate; not run here.*
## Coverage
 
- Current line coverage: **64.49%** (36,342 / 56,356 lines)
- Delta vs `main`: **+3.72 pp** (`main` baseline 60.77%, 31,844 / 52,401) — a net coverage increase from the new authority / terminality / property tests
- Coverage workstream, if applicable: n/a (authority/terminality logic; new tests added alongside)
- Exclusions added: none (no source-line coverage exclusions). *The CI coverage run excludes two daemon-spawn/kill test binaries from instrumentation — `named_group_join_metadata_event` and `x0x_0041_synthetic_kill_restart` — a pre-existing tooling detail, not a source exclusion; both still run under the normal test jobs.*
## Test Quality Checklist
 
- [x] Each new test has a why-named name stating the invariant/regression (e.g. `last_admin_gossip_apply_rejects_crafted_zero_admin_commit`, `validate_apply_rejects_live_withdrawal_without_terminal_context`).
- [x] No new test is a tautology — tests craft validly-signed, correctly-chained commits, so rejections are attributable to the authority/invariant logic, not signature/chain failure; property tests run against independent oracles.
- [x] Each new test checks one business invariant.
- [x] Assertion failures carry enough context to diagnose the broken invariant.
- [x] Coverage delta reported — **+3.72 pp** vs `main` (see Coverage).
- [x] Exclusions: none.
---
 
## Background & scope (#107 / ADR-0016)
 
This is the first phase of the plan laid out on **#107**, implementing the decisions in the accepted **[ADR-0016](https://github.com/saorsa-labs/x0x/blob/main/docs/adr/0016-role-based-group-authority-flat-admin.md)** (role-based group authority — flat Admin/Member, retiring `Owner`). It covers #107 **(a)** (retire creator/owner authority → role-based) and **(c)** (leave/delete split + last-admin invariant). #107 **(b)** (KeyPackage distribution in `MemberAdded`) is **Phase 2** and is deliberately *not* in this PR — see Deferred.
 
## How this maps to ADR-0016
 
| ADR-0016 decision | This PR |
|---|---|
| **Authority by role on the committed roster** (`ctx.members_v2` in `validate_apply`), never by `GroupInfo.creator` | All creator gates, the receive-path `sender == creator` checks, and the two-tier role matrix are deleted and replaced with the committed-roster role lookup. |
| **Retire `Owner` → flat `Admin` + `Member`; every `OwnerOnly` act → `AdminOrHigher`; `require_owner` deleted; Admin is root** | `OwnerOnly`/`require_owner` gone. Genesis seeds the creator as first Admin. Legacy `Owner` is evaluated Admin-equivalent (deserialize-compatible) but not assignable. |
| **Last-admin invariant** | `enforce_last_admin_invariant` at the state-commit choke-point, enforced apply-side on **every** path (REST + gossip) plus friendly REST pre-checks; delete is the exempt exit valve. |
| **Reserved roles (`Moderator`/`Guest`) non-assignable; role API accepts exactly `admin`/`member`** | REST role-assignment rejects `owner`/`moderator`/`guest` with explicit errors. (Apply-side acceptance discussed below — worth a look.) |
| **Phased; each phase independently shippable** (#107) | This is Phase 1. #107(b) is Phase 2; deterministic committer + race mitigations are Phase 3. |
 
## What changed
 
**Authority model.** Authority = committed-roster role + the signed commit, evaluated in `validate_apply`. Creator-identity gates removed across REST and gossip-apply. `Owner`/`OwnerOnly`/`require_owner` retired → `AdminOrHigher`.
 
**Leave / delete split.** `DELETE /groups/:id` is now a **pure self-leave for any rank** (the last admin is blocked — promote a successor or delete). Ending the group is the any-admin **delete** action over the existing `POST /groups/:id/state/withdraw` (CLI: `x0x group delete`; `state-withdraw` kept as a quiet alias). The endpoint path is unchanged for compatibility.
 
**Terminal delete = keyless tombstone.** Delete marks the local `named_groups` record `withdrawn = true` and wipes **all** key material — TreeKEM live state + snapshot + journal, MLS state, and the GSS `shared_secret`. The withdrawn record is retained as a *keyless* terminality marker (the anti-resurrection guard against a stale `GroupCard`), following the existing `withdraw_group_state`. This reads ADR-0012's "a deleted secure group leaves nothing behind" as being about *key material* (no retained keys, no recoverable continuation state) — the marker holds none.
 
**Surfaces & docs.** No surface requires an owner to exist; legacy `owner` still renders readably. The GUI gates policy/state/role controls on admin+ (was owner-only). Role-assignment API/CLI restricted to `admin`/`member`. Docs/primers/API-reference updated to "Admin is root".
 
### Receive-path authority: creator gates removed; pre-existing sender-binding left intact
 
ADR-0016 decides authority by committed-roster role in `validate_apply`. On the gossip-apply path we deleted every creator-identity authority gate (`sender == creator`) and replaced it with the role check (`at_least(Admin)`) plus signed-commit validation. The membership/role receive arms also carry a pre-existing `actor == sender_hex` sender-binding — the daemon delivers membership events directly to active members, so the actor is the sender on the normal path. We left that intact, ANDed with the role authority: it is an additional anti-spoofing restriction, **not** the authority decision and not an escalation vector (authority is the signed commit + Admin role). The terminal `GroupDeleted` receive path is different: it validates `actor == commit.committed_by` and the signed Admin-authorised withdrawal commit. Whether to relax the remaining membership/role sender-binding for full relay-tolerance is a maintainer call on the delivery model — flagged, out of ADR-0016 scope.
 
### Reserved roles: enforced at authoring, accepted on apply (deliberate)
 
Role assignment accepts exactly `admin` and `member`. The **role-assignment REST API** rejects `owner`, `moderator`, and `guest` with explicit errors (per ADR-0016: reserved, non-assignable). The **signed gossip-apply path** accepts any validly-signed role label *except* `Owner`. This split is deliberate and consistent with the ADR placing role-vocabulary enforcement on the *assignment API* (apply-side enforcement is reserved for the last-admin invariant):
 
- **No authority is at stake.** The only authority threshold in the code is `role.at_least(Admin)` (ranks: Owner 4, Admin 3, Moderator 2, Member 1, Guest 0). `Moderator`/`Guest` fall below `Admin` and appear in no privilege check — an applied reserved role is a member-level participant. `Owner` is the one reserved role that *is* admin-equivalent, so it remains rejected on apply.
- **Only an admin can author a role-update**, and an admin can already add anyone as a member directly — so a signed commit carrying `Moderator`/`Guest` grants no capability its author didn't already have.
- **Apply must accept validly-signed peer commits** — including from older/legacy daemons — or replicas diverge and the group forks across versions. Vocabulary restrictions therefore belong at *assignment* (authoring), not *receipt* (apply); enforcing them on apply would break byte-for-byte legacy replay and fork live state, for no security gain.
This matches the "Admin is root" model: a non-admin label carries no authority. The membership tests assert this accept-on-apply behaviour intentionally.
 
### Delete / terminal withdrawal is admin-only at the authority choke-point
 
`x0x group delete` / `POST /groups/:id/state/withdraw` is the explicit group-ending action — admin-only at both the REST authoring path and signed-commit apply. `validate_apply` is **fail-closed by default**: an ordinary apply rejects any live→withdrawn commit (`"live withdrawal commit requires explicit terminal validation"`); only `validate_apply_terminal` (the `GroupDeleted` path) permits it, and only for an Admin/legacy-Owner signer. So a non-admin self-leave — or any non-terminal action — cannot terminally withdraw the group by smuggling the flag through (e.g.) `MemberRemoved`. The flag and the key-wipe are coupled: a group becomes withdrawn only via the path that also wipes key material, leaving a coherent keyless tombstone. Already-withdrawn state stays terminal (un-withdrawing is rejected).
 
### A behaviour change to the existing card-import path (flagged for confirmation)
 
The withdrawn-`GroupCard` import path previously marked/closed a live local group on the card's revision + signature alone. With key-wipe now attached to withdrawal, that would become destructive on a bare card, so we tightened it: a **live, keyed** group can no longer be terminally withdrawn by a bare card — that requires a chain-validated signed withdrawal commit (the `GroupDeleted` path). A bare withdrawn card may still supersede a *keyless discovery stub*. This is the one place the change alters existing behaviour; the prior behaviour can be preserved instead if the maintainer prefers.
 
## No wire / format change (verified)
 
Diffed against `main`. **Zero** change to: the signed `GroupStateCommit` structure, its canonical byte encoding, signing, or `verify_structure`; `compute_roster_root` / `compute_state_hash`; the `GroupRole` wire encoding (`as_u8`), serde, ordering, or variants; the `NamedGroupMetadataEvent` gossip enum; or the on-disk storage schema. Legacy chains verify byte-for-byte. The retired `OwnerOnly` is an internal, non-serialized enum. No dependency changes (`Cargo.toml`/`Cargo.lock` untouched).
 
**Scope of this claim:** this is **wire / hash / storage** compatibility — *not* a claim that an old (pre-Phase-1) daemon will accept every new Phase-1 authority action during mixed-version operation. An old daemon still running creator/owner gates may reject a new non-creator-Admin action; that is inherent to changing the authority model and resolves once peers are on Phase 1.
 
## Backward compatibility & downstream impact
 
No data/key migration is needed (legacy `Owner` records and chains are handled automatically). There are, however, **API-behaviour** changes any daemon client should know about:
 
1. **`DELETE /groups/:id` is now "leave a group" (self-leave), not delete/disband.** A client that used `DELETE` to *end* a group should switch to `POST /groups/:id/state/withdraw` (`x0x group delete`).
2. **Role assignment rejects `owner`/`moderator`/`guest`** (admin/member only) with explicit errors.
3. **New last-admin `409`s** on leave/remove/ban/demote of the last admin.
4. **Deleted groups now linger in `GET /groups` as keyless tombstones** (a creator-`DELETE` previously removed the record). The list response doesn't currently expose a `withdrawn` flag, so a list consumer should filter on the group's `/state` (or `withdrawn` could be added to the list response as a small additive follow-up).
Downstream: the **in-repo GUI is updated in this PR** (controls re-gated owner→admin+, "Delete (end for everyone)", uses the `/state/withdraw` endpoint). **Communitas** is external — if it does group management it should check items 1–4 (its usage was not verifiable from this repo). **Symphony** appears unaffected — `docs/symphony-integration.md` references none of the changed surfaces.
 
## Test hygiene
 
This PR removes a committed `audit.jsonl` artifact and `.gitignore`s it, and adjusts one exec test helper to route a relative audit-log path into a tempdir so test runs don't write a tracked file. These are the only changes outside the group-authority/terminality scope and carry no runtime change — flagged so they don't puzzle on review; can be split into a separate commit if preferred.
 
## Deferred to later phases (per #107 — openly stated, not missed)
 
- **KeyPackage distribution in `MemberAdded` is Phase 2 (#107 b).** It touches the signed commit format, so — per the process set out on #107 — the **wire shape should be agreed on the #107 thread before implementation**; we have not pre-empted it. Consequences in the interim: **delegated (non-creator-admin TreeKEM) ban is non-operational** (the existing "member is missing TreeKEM KeyPackage" / 424 boundary is unchanged); and a card-derived stub's join-approval can't converge requester-side until roster/KeyPackage propagation lands.
- **Deterministic committer, rebase-and-retry, race mitigations, and self-leave PCS rekey are Phase 3.** The two-admin metadata race window therefore exists in the interim (exposed, not solved, per the ADR).
## Diff breakdown — where the +10k goes
 
The raw `+10,174 / −974` is large; here's where it actually lands. The work splits roughly in half:
 
| Bucket | Production | Tests | What it is |
|---|---|---|---|
| **Authority / roles (ADR-0016)** | **net subtraction** — gates + `OwnerOnly` + `require_owner` removed (35 lines, 0 added); only the small last-admin invariant added | ~half | The point of the PR — the authority model gets *simpler*. |
| **Keyless-tombstone delete subsystem (ADR-0012)** | **≈ +1,500** | ~half | Brings delete up to ADR-0012 consistently + anti-reanimation + in-flight race safety (see below). |
| Docs | ≈ +80 | — | API-reference, primers, ADR. |
 
Two things to pull out: the **shipped/production code grows by only ~1,900 lines** (the rest, ~80%, is tests), and that growth is **entirely the ADR-0012 delete subsystem** — the ADR-0016 authority change itself is a net deletion. The ~7,300 lines of tests divide fairly evenly between proving the authority rules and proving the delete subsystem.
 
### Why the delete subsystem exists
 
It isn't surface bolted onto the authority change — it closes pre-existing gaps in the delete paths, which are inconsistent in the current code:
 
- The **`withdraw` path** (which the new admin "delete" builds on) seals the group withdrawn but **wipes no key material** — `shared_secret`, TreeKEM, MLS, snapshot and journal all survive. An ADR-0012 gap.
- The **creator-`DELETE` path** *does* wipe (it even carries an explicit `// ADR-0012` comment) but removes the group's record **entirely**, leaving no tombstone — so a stale `GroupCard` can resurrect a "deleted" group.
This PR unifies them into one correct delete: **wipe all key material** (closing the withdraw-path leak, honouring ADR-0012 the way the creator-delete path already intends) **+ keep a keyless tombstone** (closing the reanimation hole that full-record removal leaves) **+ recheck against in-flight operations** that could re-install material mid-delete (a race the simpler delete is also exposed to). The bulk of the subsystem, and most of its tests, is that in-flight robustness.
 
### If delete scope is a question
 
ADR-0012 itself isn't really in question — the creator-delete path already implements it. But the *robust, consistent* version of it is roughly **half of this PR** (~+1,500 production plus its tests), so that's where the size impact of any change in delete scope would sit. The one genuine design choice for review is the **keyless tombstone** (retain a keyless record for anti-reanimation) vs the current full-record removal — already discussed with the review team. A lighter delete would shrink the diff by about that half, at the cost of re-opening the leak, the reanimation hole, or the in-flight races this closes.
 
### On the test volume
 
~80% of the diff is tests, split fairly evenly between the authority rules and the delete subsystem; they lift line coverage +3.72pp. The `tests/*.rs` files are independent (droppable per file), and the heavier ones can be tiered behind `#[ignore]` / a separate profile (as the Soak Test already is) if a lighter regular-CI footprint is wanted — though they're fast deterministic regression guards, so the default is to keep them.
 
## Where review attention is most valuable
 
A few points below are maintainer judgement calls — flagging for @dirvine:
 
1. **The authority choke-point** — `validate_apply` / `validate_apply_terminal` in `state_commit.rs`: role-based authority + the fail-closed live→withdrawn terminal rule.
2. **Reserved roles accepted on apply** — the deliberate authoring-vs-apply split above. We believe it's correct and required for convergence; it's the one decision a fresh reader most often reads as a bug, so an explicit confirmation would be valuable.
3. **The card-import behaviour change** — the one place this alters existing behaviour.
4. **ADR-0012 reading** — "leaves nothing behind" = *key material*, so a keyless withdrawn marker is retained as the anti-resurrection guard. Confirmation that this matches the intended reading would help.
5. **The verb** (following up the open naming question raised on #107) — the ADR uses "delete" and we implemented `x0x group delete` to match (endpoint path stays `/state/withdraw`). We'd gently suggest **"disband"** reads more accurately for *ending the group for everyone* (vs an individual leaving) — it's isolated to a single constant — but "delete" was shipped to match the ADR, and the call is the maintainer's.
6. **The Soak Test** — the upstream endurance gate, which we could not run.
## Maintainer follow-ups (flagged, not fixed here)
 
- The pre-existing `actor == sender_hex` sender-binding on the membership/role receive arms — left as the existing delivery-model design; whether to relax it for full relay-tolerance is a maintainer call.
- The raw `/mls/groups/:id/members` endpoints (separate pre-existing subsystem, different keyspace, wiped on delete) — can't resurrect a withdrawn named group, but worth recording.
- The discovery-card receive paths are best-effort signed-hint caches, not current-roster-admin authority (pre-existing) — a maintainer call on whether known-group imports should be hardened.
- Retention of the keyless withdrawn marker is indefinite (matching the existing `withdraw`); whether to GC/expire it is a product/privacy decision for the maintainer.
- Optional: add `withdrawn` to the `GET /groups` list response so clients can filter deleted tombstones.
- **Test-environment observation (pre-existing, not introduced here):** the multi-daemon harness tests (e.g. `named_group_join_metadata_event`) don't form a mesh on macOS locally — they fail at bring-up with "zero peers" — so those multi-daemon paths couldn't be exercised on the contributor's machine. They pass on Linux CI (Multi-Agent Integration is green, see Validation) and the failure reproduces on a clean pre-Phase-1 baseline, so it's environmental rather than a regression here. Flagging in case the local-macOS harness experience is worth a look.

<h3>Greptile Summary</h3>

Phase 1 of ADR-0016 replacing the `Owner`-authority model with a flat Admin/Member model. The core authority decision is moved from `GroupInfo.creator` identity to the signer's role on the committed roster, evaluated in `validate_apply` / `validate_apply_terminal`. The PR also closes pre-existing ADR-0012 gaps by unifying delete into a keyless tombstone path that wipes all key material.

- **Authority model**: `OwnerOnly`/`require_owner` removed; every privileged action gated on `AdminOrHigher` (legacy `Owner` entries evaluate as Admin via `at_least`). A last-admin invariant (`enforce_last_admin_invariant`) is enforced at both authoring (`seal_commit`) and apply (`finalize_applied_commit`) choke-points, covering REST, gossip, and TreeKEM paths.
- **Leave/delete split**: `DELETE /groups/:id` is now a pure self-leave; terminal group deletion moves to `POST /groups/:id/state/withdraw` (new CLI: `x0x group delete`). A withdrawal commit produces a keyless tombstone (retained for anti-resurrection) with all key material wiped.
- **Fail-closed terminal gate**: `validate_apply` rejects live→withdrawn commits; only `validate_apply_terminal` (reached exclusively via `GroupDeleted` events) permits them. `store_named_group_info_locked` provides a second anti-resurrection guard, preventing non-withdrawn state from overwriting a withdrawn tombstone.

This PR is safe to merge; the authority model and terminal-delete subsystem are correctly implemented with no key-material leakage paths on the happy path.

The core authority invariant is enforced at every relevant choke-point — commit authoring, gossip apply, REST pre-check — and the fail-closed validate_apply / validate_apply_terminal split reliably gates live→withdrawn transitions. The three comments posted are style-level: assignable_from_name misleads mis-cased callers, a redundant clear_group_info_key_material call should be removed for clarity, and a narrow TOCTOU window in the TreeKEM join install path is already deferred to Phase 3. None of these affect correctness on any currently reachable code path.

The src/server/mod.rs double-check pattern in install_joined_treekem_group_after_crypto_recheck warrants a Phase 3 revisit. Everything else is well-covered by the new test suite.

| Filename | Overview |
|----------|----------|
| src/groups/state_commit.rs | Core authority choke-point: OwnerOnly removed, validate_apply / validate_apply_terminal split correctly enforces the fail-closed live→withdrawn gate; enforce_last_admin_invariant added and well-tested. |
| src/groups/mod.rs | GroupInfo authoring path updated: seal_withdrawal adds admin check + rollback; seal_commit gains last-admin invariant guard; finalize_applied_terminal_commit added as explicit terminal finalizer path. |
| src/server/mod.rs | Large but well-structured: REST handlers uniformly adopt require_admin_or_above + reject_withdrawn_group + last_admin_precheck; GroupDeleted receive path correctly routes through validate_apply_terminal; retain_withdrawn_group_tombstone and wipe_local_group_crypto_material implement key-wipe-then-tombstone correctly. Double-check TOCTOU window in install_joined_treekem_group_after_crypto_recheck is acknowledged as Phase 3 deferred work. |
| src/groups/member.rs | new_admin replaces new_owner for group creation; assignable_from_name enforces lowercase-only admin/member vocabulary at REST assignment; new_owner kept for legacy compatibility. |
| src/groups/invite.rs | creator_agent_id_from_base_state is a best-effort heuristic for display-only creator provenance; correctly documented as non-authority and non-tamper-evident. |
| tests/last_admin_invariant.rs | 362 lines of dedicated invariant tests including gossip-apply rejects for crafted zero-admin commits; well-named and independently verifiable. |
| tests/membership_authority.rs | 893 lines covering every authority boundary (REST + gossip) including accept-on-apply of reserved Moderator/Guest roles; explicit assertions on the authoring-vs-receipt split. |
| tests/owner_retirement.rs | 429 lines verifying Owner-equivalent authority, legacy-chain replay byte-for-byte compatibility, and the Owner→Admin normalisation path. |
| tests/proptest_groups.rs | Property-based tests for the authority model; oracles independently verify invariants across arbitrary roster shapes. |
| src/cli/commands/group.rs | New x0x group delete maps to POST /state/withdraw; state_withdraw kept as alias; DELETE_VERB constant isolates the maintainer-facing verb-swap handle cleanly. |

</details>

<h3>Flowchart</h3>

<a href="#gh-light-mode-only">

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
    A[REST: POST /groups/:id/state/withdraw] --> B{require_admin_or_above}
    B -->|fail| ERR1[403 Forbidden]
    B -->|pass| C{reject_withdrawn_group}
    C -->|fail| ERR2[409 Conflict]
    C -->|pass| D[info.seal_withdrawal]
    D --> E[shared_secret = None]
    E --> F[retain_withdrawn_group_tombstone]
    F --> G[wipe_local_group_crypto_material]
    F --> H[save_named_groups]
    F --> I[emit GroupDeleted event]

    J[Gossip: GroupDeleted event] --> K{pre-filter checks}
    K -->|fail| DROP[silently drop]
    K -->|pass| L[apply_terminal_stateful_event_to_group]
    L --> M[validate_apply_terminal]
    M -->|fail| DROP
    M -->|pass| N[finalize_applied_terminal_commit]
    N --> O[retain_withdrawn_group_tombstone]

    P[REST: DELETE /groups/:id] --> Q{last_admin_self_leave_precheck}
    Q -->|last admin| ERR3[409 – promote successor first]
    Q -->|ok| R[MemberRemoved self-leave]

    style ERR1 fill:#f96
    style ERR2 fill:#f96
    style ERR3 fill:#f96
    style DROP fill:#ffc
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%%
flowchart TD
    A[REST: POST /groups/:id/state/withdraw] --> B{require_admin_or_above}
    B -->|fail| ERR1[403 Forbidden]
    B -->|pass| C{reject_withdrawn_group}
    C -->|fail| ERR2[409 Conflict]
    C -->|pass| D[info.seal_withdrawal]
    D --> E[shared_secret = None]
    E --> F[retain_withdrawn_group_tombstone]
    F --> G[wipe_local_group_crypto_material]
    F --> H[save_named_groups]
    F --> I[emit GroupDeleted event]

    J[Gossip: GroupDeleted event] --> K{pre-filter checks}
    K -->|fail| DROP[silently drop]
    K -->|pass| L[apply_terminal_stateful_event_to_group]
    L --> M[validate_apply_terminal]
    M -->|fail| DROP
    M -->|pass| N[finalize_applied_terminal_commit]
    N --> O[retain_withdrawn_group_tombstone]

    P[REST: DELETE /groups/:id] --> Q{last_admin_self_leave_precheck}
    Q -->|last admin| ERR3[409 – promote successor first]
    Q -->|ok| R[MemberRemoved self-leave]

    style ERR1 fill:#f96
    style ERR2 fill:#f96
    style ERR3 fill:#f96
    style DROP fill:#ffc
```

</a>

`````markdown

---

### Issue 1 of 3
src/groups/member.rs:90-99
The `assignable_from_name` matcher treats any string that isn't lowercase `"admin"` or `"member"` as a "reserved" role, including mis-cased variants like `"Admin"` or `"ADMIN"`. The resulting error message — `"role 'ADMIN' is reserved and cannot be assigned"` — is misleading: the caller has the right intent but the wrong casing. Consider folding the input to lowercase before matching, or at minimum issuing a distinct "role names are case-sensitive" message for mis-cased versions of valid names.

```suggestion
    pub fn assignable_from_name(name: &str) -> Result<Self, String> {
        match name {
            "admin" => Ok(Self::Admin),
            "member" => Ok(Self::Member),
            "owner" => Err(Self::OWNER_ASSIGNMENT_ERROR.to_string()),
            other => {
                if other.eq_ignore_ascii_case("admin") || other.eq_ignore_ascii_case("member") {
                    return Err(format!(
                        "role names are case-sensitive; use '{}'",
                        other.to_lowercase()
                    ));
                }
                Err(format!(
                    "role '{other}' is reserved and cannot be assigned; valid roles: 'admin', 'member'"
                ))
            }
        }
    }
```

### Issue 2 of 3
src/server/mod.rs:7728-7760
**Residual TOCTOU window after second terminality recheck**

`install_joined_treekem_group_after_crypto_recheck` calls `ensure_named_group_key_material_install_allowed` a second time (after `persist_treekem_and_named_groups_atomic_with_info`) before inserting into `treekem_groups`. If a concurrent withdrawal completes and wipes `treekem_groups` between the second check and the `treekem_groups.write().await.insert(...)`, the key material is installed into the in-memory map after the wipe, leaving it resident until the next restart or eviction. The PR description explicitly defers this to Phase 3; flagging here for tracking. The double-check already substantially narrows the window vs the prior design.

### Issue 3 of 3
src/server/mod.rs:12983-12996
`seal_withdrawal` already clears `info.shared_secret` to `None` on success — that is part of its documented contract since this PR. The explicit `clear_group_info_key_material(info)` call here is therefore redundant. Removing it prevents future readers from wondering whether `seal_withdrawal` silently omits the clear in some paths, and keeps the responsibility unambiguously inside the library method.

```suggestion
        let commit = match info.seal_withdrawal(signing_kp, now_ms) {
            Ok(c) => c,
            Err(e) => {
                return api_error(
                    StatusCode::INTERNAL_SERVER_ERROR,
                    format!("withdrawal seal failed: {e}"),
                );
            }
        };
        // Delete retains a keyless withdrawn tombstone: ADR-0012's "leave
        // nothing behind" means no MLS/TreeKEM/GSS key material survives, not
        // that the terminal metadata record is deleted. Keeping this record is
        // the stale-card reanimation guard for future imports.
        // Note: seal_withdrawal already clears shared_secret on success.
```

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["docs(adr-0016): use durable reserved-rol..."](https://github.com/saorsa-labs/x0x/commit/514feb8319bbed2acb0403f3f3a90de1321ecae3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=39799815)</sub>


dirvine (commented):
# Local Review — ADR-0016 Phase 1 (flat Admin/Member authority)

Full local review of `feat/adr-0016-phase-1-on-v0.26` @ `514feb8` — cross-validated by three independent reviewers (Codex, Greptile, and a 57-agent workflow review) plus a manual read and a full local build/test run.

## Local validation (clean worktree, crates.io deps)

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | ✅ pass |
| `cargo clippy --all-targets --all-features -- -D warnings` | ✅ pass (clean) |
| `cargo nextest run --all-features --workspace` | ⚠️ 1884 passed, **1 failed**, 190 skipped |
| Production `unwrap`/`expect`/`panic!` introduced | ✅ none (all in `#[cfg(test)]`) |

The single failure — `non_creator_admin_private_secure_invite_e2e_converges_through_real_daemons` — **passes in isolation** (mesh forms, all 3 daemons connect, converges in ~41s). It's the known macOS loopback-mesh flake under port/resource contention with ~1886 concurrent tests, exactly as the PR flagged. CI Linux is green. **Not a logic regression.**

## What's solid

The core authority change is genuinely well-built, and the highest-risk areas are cleared:

- **No authority bypass.** `validate_apply` / `validate_apply_terminal` decide on committed-roster role (`AdminOrHigher`), never on `creator`. Terminal withdrawal is *double-gated*: requires both terminal mode **and** an active Admin signer (`state_commit.rs:109-136`).
- **Last-admin invariant** enforced at both choke-points — authoring (`seal_commit`, `mod.rs:500`) and apply (`finalize_applied_commit`, `mod.rs:670`) — covering REST + gossip + TreeKEM. Well-tested (demote/remove/ban/self-leave).
- **Key-wipe coupled to withdrawal.** `seal_withdrawal` rolls back cleanly on seal failure; the tombstone is inserted *before* the wipe, and the anti-resurrection guard (`store_named_group_info_locked`) refuses non-withdrawn overwrites.
- Net-simpler authority code; no wire/hash/storage change.

## Findings — real bugs (not intended)

### 🔴 HIGH — Agent-card invites are unjoinable in this build (`server/mod.rs:4398` + `:11282`)
The new join path hard-rejects (`bad_request`, 400) when `invite.creator_agent_id_from_base_state()` errors, which it does whenever `base_members_v2` is absent. The proper generate-invite endpoint (`:11101`) sets `base_members_v2` (`:11114`), but the **agent-card path** (`GET /agent/card?include_groups=true`, line 4398) builds `SignedInvite::new()` and never sets `base_members_v2` / `base_state_revision` / `group_created_at`. So any invite shared via an agent card — generated by *this same build* — returns `"invite missing base member snapshot; cannot derive creator provenance"` on join.

The e2e tests only exercise the generate-invite endpoint, so they miss this path. **Recommend fixing before merge:** set the base-state fields at `:4398` (mirroring `:11101-11114`), and add a join-path test over a card-derived invite.

### 🟠 MEDIUM — TOCTOU: TreeKEM key material can be re-installed after a withdrawal wins the race (`server/mod.rs:7716`)
`install_joined_treekem_group_after_crypto_recheck` rechecks the tombstone, then inserts into `treekem_groups` *outside* that critical section. A concurrent `retain_withdrawn_group_tombstone` landing between the recheck and the insert leaves `named_groups` withdrawn/keyless but `treekem_groups` holding live keys. (Flagged by Codex as High and by Greptile inline; the PR explicitly defers it to Phase 3.)

Assessment: real but bounded — in-memory only (the persisted snapshot is already wiped; residue clears on restart) and needs precise interleaving. The fix is small (final withdrawn-check while holding the `treekem_groups` write lock, bail+remove if withdrawn) and **worth doing now** rather than deferring, since it's a key-residency invariant.

### 🟠 MEDIUM — GUI: admin demotion control removed (`gui/x0x-gui.html:4462`)
Only "Promote to admin" remains; there is no way to demote an admin back to member. Combined with the last-admin rule ("promote a successor before leaving"), an over-promoted roster can't be walked back from the GUI.

### 🟡 LOW — GUI: caller's own row not excluded from Ban/Promote (`gui/x0x-gui.html:4456`)
An admin can click Ban on their own row.

### 🟡 LOW — Withdrawn card leaks in discovery via stable-id alias (`server/mod.rs:718`)
`remove_group_card_if_not_stale` drops only under `card.group_id`; the `stable_group_id` alias entry (written by `refresh_group_card_cache_from_info`, `:7612-7613`) survives until TTL, so a deleted group can stay discoverable.

### 🟡 LOW — Mixed-version roster divergence (`groups/mod.rs:674`)
A legacy peer's validly-signed last-admin-stripping commit is silently dropped by new peers, forking the chain. Inherent to the authority change; the PR acknowledges mixed-version caveats.

## Intended-but-breaking — please ratify against ADR-0016 (do not "fix")

These are the *point* of the PR and consistent with "Admin is root" — flagging so they're conscious decisions, not surprises:
- **Any admin can delete the group** (`:8984`) — creator loses exclusive veto over destruction.
- **Any admin can remove the original creator** via the member API (`:11877`) — creator-only protection gone.
- **`DELETE /groups/:id` is now self-leave, not delete** (`:13047`) — clients that used it to *end* a group must move to `POST /state/withdraw`.
- **Creator leaving a TreeKEM group is now an ordinary self-leave** (`:11965`, `CreatorMustDelete` removed).

## Cleanups (optional)
- `assignable_from_name` gives a misleading "reserved" error for mis-cased `ADMIN`/`MEMBER` (Greptile #1).
- Redundant `clear_group_info_key_material` after `seal_withdrawal` already nulls `shared_secret` (`:12996`, Greptile #3); conversely 3 sites bypass the wrapper with direct `shared_secret = None` (`:9157` / `:10317` / `:11192`) — pick one convention.
- `last_admin_precheck_error` deep-clones all of `GroupInfo` to test a `members_v2`+`withdrawn` invariant (`mod.rs:268`).
- Copy-pasted stale-card-drop block in 3 listeners (`~:6792` / `:7073` / `:7306`).
- `tests/CLAUDE.md` integration table not updated with the 4 new test files (`owner_retirement`, `membership_authority`, `last_admin_invariant`, `invite_authority`).
- **Rule 9 concern:** `tests/membership_authority.rs` re-implements handler authority logic rather than invoking the handlers — so it would *not* catch a dropped `require_admin_or_above` in a REST handler, which is precisely the class of change this PR makes.

## Verdict

**Strong PR — the authority model and terminal-delete subsystem are correct and exceptionally well-tested.** One blocking functional regression (agent-card invites unjoinable) and one should-fix-now key-residency TOCTOU stand between this and merge; the two GUI gaps are merge-soon. Everything else is ratification or polish.

**Recommendation: request changes** — fix the agent-card invite path + add a covering test, apply the small TOCTOU lock fix, restore GUI demote / self-row exclusion — then merge.

---
*Reviewed locally with `cargo fmt`/`clippy`/`nextest` + Codex (`codex-cli 0.133.0`) + a 57-agent workflow review. Authority logic, choke-points, key-wipe coupling, and the no-new-panics rule were each manually verified against the diff.*

dirvine:
# PR #121 — Review-team cross-check (reconciled)

I cross-checked the 5-reviewer synthesis against the latest local review already on the PR. The local review surfaced additional findings outside the original reviewer scopes; I verified the important ones directly against source at `514feb8` and incorporated them below.

**Reconciled verdict: request changes.** The authority model and terminal-delete subsystem are correct and well-tested, but the local review's blocking functional regression (agent-card invites) and one key-residency TOCTOU stand between this and merge.

Validation reported by reviewers/local review is clean for fmt/clippy/check; the latest full local `nextest` had one known macOS loopback-mesh flake (`non_creator_admin_private_secure_invite_e2e_converges_through_real_daemons`) that passes in isolation; CI Linux is green.

---

## Where the cross-check changed the picture

Both reviews agree on the core: authority choke-points are sound, the last-admin invariant is correctly placed at both choke-points, key-wipe is coupled to withdrawal, the wire format is unchanged, and no production `unwrap`/`expect`/`panic!` was introduced. The local review surfaced three findings outside the original 5-reviewer scopes; each verified against source:

### 🔴 HIGH (blocking) — Agent-card invites are unjoinable — confirmed
`GET /agent/card?include_groups=true` (`src/server/mod.rs:4390–4406`) builds `SignedInvite::new()` and never sets `base_members_v2` / `base_state_revision` / `group_created_at` — none of the 11 base-state fields the generate-invite path populates (`:11101–11120`). On join, `creator_agent_id_from_base_state()` (`src/groups/invite.rs:225`) errors when `base_members_v2` is `None`, and the join handler returns `bad_request` (`src/server/mod.rs:11282–11286`). Any invite shared via an agent card from this build returns `"invite missing base member snapshot; cannot derive creator provenance"`. The e2e tests only exercise the generate-invite endpoint, so they miss this path.

**Fix:** set the base-state fields at `:4390–4406` mirroring `:11101–11120`, and add a join-path test over a card-derived invite.

### 🟠 MEDIUM — TreeKEM install TOCTOU — confirmed, worth fixing now
`install_joined_treekem_group_after_crypto_recheck` (`src/server/mod.rs:7716`) rechecks the tombstone, then inserts into `treekem_groups` outside that critical section. A concurrent withdrawal landing between recheck and insert leaves `named_groups` withdrawn/keyless but `treekem_groups` holding live keys. Impact is in-memory only (persisted snapshot already wiped; residue clears on restart) and needs precise interleaving, but it violates a key-residency invariant and the fix is small: a final withdrawn-check while holding the `treekem_groups` write lock, bail+remove if withdrawn. Suggest doing this now rather than deferring to Phase 3.

### 🟠 MEDIUM + 🟡 LOW — GUI gaps — confirmed
- **No demote control** (`src/gui/x0x-gui.html:4468`): only a "Promote to admin" button exists. Combined with the last-admin rule, an over-promoted roster can't be walked back from the GUI.
- **Caller's own row not excluded from Ban/Promote**: an admin can click Ban on their own row. No self-row guard found in roster rendering.

### Rule 9 test-architecture note — confirmed, sharp
`tests/membership_authority.rs` documents that it *"mirror[s] the daemon's role pre-check"* — it re-implements handler authority logic rather than invoking the real REST handlers. The tests are correct at their layer (and the proptest oracles are genuinely independent), but they test a mirror of the handler, not the handler itself. A future PR that drops a `require_admin_or_above` from a REST handler would not be caught — exactly the class of change this PR's future follow-ups risk. The daemon-spawning tests that do exercise real endpoints are `#[ignore]`'d. Worth at least one end-to-end REST authority test (or un-ignoring a daemon test) for this class of regression.

---

## Load-bearing security claims — verified independently

Both reviews and my own source check agree on these:

1. **`validate_apply` is fail-closed against live→withdrawn** (`state_commit.rs:754–787`). Only `validate_apply_terminal` (solely via `GroupDeleted`) permits it, and only for `AdminOrHigher`.
2. **Authority comes from the committed roster, not the incoming commit** — `apply_commit` passes `members_v2: &self.members_v2` into `ApplyContext`; an attacker cannot craft a commit granting itself Admin.
3. **Key wipe is atomic with the withdrawal marker** (`groups/mod.rs:675`) — `shared_secret = None` + `wipe_local_group_crypto_material` + tombstone persisted before wipe + `store_named_group_info_locked` anti-resurrection guard.
4. **Wire format is byte-identical** — `GroupRole` enum, `as_u8` discriminants (0–4), `serde(rename_all="snake_case")`, `compute_roster_root`, `compute_state_hash`, `GroupStateCommit`, `NamedGroupMetadataEvent` all unchanged vs `main`. Legacy chains verify byte-for-byte (`owner_retirement_legacy_owner_chain_replays_byte_for_byte` pins a blake3 hash proving this).
5. **Raw-withdrawal bypass closed** — pre-PR `validate_apply` accepted live→withdrawn from any `AdminOrHigher`; now only `validate_apply_terminal` permits it.

---

## Consolidated, de-duplicated finding list

### 🔴 Blocking
1. **Agent-card invites unjoinable** — set base-state fields at `server/mod.rs:4390–4406`; add covering join test.

### 🟠 Should-fix-now (before merge)
2. **TreeKEM install TOCTOU** — final withdrawn-check under the `treekem_groups` write lock; bail+remove if withdrawn.
3. **GUI: no demote control** (`gui/x0x-gui.html:4468`) — restore a demote-to-member affordance.
4. **GUI: caller's own row not excluded from Ban/Promote** — add self-row guard.

### 🟡 Worth tracking (non-blocking)
5. **Withdrawn card leaks in discovery via stable-id alias** (`server/mod.rs:718`) — `remove_group_card_if_not_stale` only checks `card.group_id`, not the `stable_group_id` alias written by `refresh_group_card_cache_from_info`.
6. **Mixed-version roster divergence** (`groups/mod.rs:674`) — legacy peer's last-admin-stripping commit silently dropped by new peers, forking the chain. Inherent to the authority change; document the upgrade caveat (one-line note in v0.26 CHANGELOG / ADR-0016 migration section: upgrade all members before authoring new commits).
7. **Rule 9: membership tests mirror handler logic** — add at least one end-to-end REST authority test (or un-ignore a daemon test) so a dropped `require_admin_or_above` would be caught.
8. **GET /groups tombstone gap undocumented** — list response has no `withdrawn` field; clients silently receive tombstones. Document the `/groups/:id/state` workaround or add the flag as a fast-follow.
9. **Exact error strings missing from REST docs** — last-admin 409, role-rejection (owner/moderator/guest); API clients can't pattern-match.
10. **Add gossip-apply `BanMember`-of-sole-admin test** — the one gap in the last-admin apply-path matrix (`MemberBanned` is a distinct event arm; demotion apply-path tests already exist at `tests/last_admin_invariant.rs:231`/`:253`, only the ban apply-path case is missing).

### ⚪ Cleanups (optional)
- `assignable_from_name` misleading "reserved" error for mis-cased `ADMIN`/`MEMBER` (Greptile #1).
- Redundant `clear_group_info_key_material` after `seal_withdrawal` (`:12996`); conversely 3 sites bypass the wrapper with direct `shared_secret = None` (`:9157`/`:10317`/`:11192`) — pick one convention.
- `last_admin_precheck_error` deep-clones all of `GroupInfo` to test a `members_v2`+`withdrawn` invariant (`mod.rs:268`).
- Copy-pasted stale-card-drop block in 3 listeners (`~:6792`/`:7073`/`:7306`).
- `tests/CLAUDE.md` integration table not updated with the 4 new test files.
- `DELETE_VERB` constant; `new_owner()` test-only gating; `src/server/mod.rs` delete-subsystem module organization; negative-test assertion specificity.

### Intended-but-breaking — ratify against ADR-0016 (do NOT "fix")
These are the point of the PR per "Admin is root." Flagging so they're conscious decisions: any admin can delete the group (`:8984`); any admin can remove the original creator (`:11877`); `DELETE /groups/:id` is now self-leave (`:13047`); creator leaving a TreeKEM group is an ordinary self-leave (`:11965`, `CreatorMustDelete` removed).

---

## Bottom line

The authority model and terminal-delete subsystem are correct, exceptionally well-tested at the library layer, and wire-compatible. The PR is not ship-ready as-is: the agent-card invite regression is blocking, the TreeKEM install TOCTOU violates a key-residency invariant with a small available fix, and the two GUI gaps are user-facing. The Rule 9 test-mirror concern is the most important methodological note for future follow-ups.

**Recommendation: request changes** — fix #1 (blocking) and #2–#4 (should-fix-now), then merge.

— *5-reviewer synthesis (security/crypto · authority · tests · simplicity · docs/API), cross-checked against the latest local review. All cross-review findings re-verified against source at `514feb8` before posting.*

dirvine (changes requested):
Requesting changes based on the [review-team cross-check](https://github.com/saorsa-labs/x0x/pull/121#issuecomment-4800229814).

**Blocking:** agent-card invites unjoinable (`src/server/mod.rs:4390–4406` never sets `base_members_v2` / base-state fields; join returns `bad_request`).

**Should-fix-now:** TreeKEM install TOCTOU (`:7716` — final withdrawn-check under the `treekem_groups` write lock); GUI demote control missing (`gui/x0x-gui.html:4468`); GUI self-row not excluded from Ban/Promote.

Full detail in the linked comment. Authority model, terminal-delete subsystem, wire-format compat, and the load-bearing security claims all verified sound — these are the fixes between this and merge.

JimCollinson:
Thanks for the depth here! The cross-checked read is exactly the scrutiny I was hoping for on this, especially on the authority and delete paths, and it's reassuring the core landed sound. Agreed with the verdict; here's how we'll close it out.

**Ratifying the intended-but-breaking four (not fixing).** These are deliberate and exactly the ADR-0016 "Admin is root" model, confirming so they're on the record as conscious decisions:

- Any admin can delete the group — destruction is an ordinary admin act; the creator's exclusive veto is intentionally gone.
- Any admin can remove the original creator — creator is history, not standing authority.
- `DELETE /groups/:id` is self-leave, not delete — the leave/delete split; ending the group is the any-admin withdraw.
- Creator leaving a TreeKEM group is an ordinary self-leave (`CreatorMustDelete` removed) — same principle.

All four are ADR-0016 as accepted. No change intended.

**Fixing before merge:**

- **Blocking — agent-card invites unjoinable.** Confirmed on our side too: the agent-card invite builder (`/agent/card?include_groups=true`) never sets the base-state fields the new provenance-deriving join path requires, while generate-invite does — so card-shared invites get rejected. We'll populate those fields on the card path (mirroring generate-invite) and add a join test over a *card-derived* invite. We're leaving the join-path check as-is — it's the authority hardening; the bug is the emitter, not the gate.
- **TOCTOU (treekem install).** Fair call to pull it forward — we'll do the final withdrawn-check while holding the `treekem_groups` write lock and bail+remove if withdrawn, rather than defer the window to Phase 3.
- **GUI demote + self-row.** Adding a demote-to-member control (admin-gated, last-admin-aware) and excluding the caller's own row from Ban/Promote.

**On the Rule 9 point — agreed, and it's the most useful note in the review.** The membership tests assert against a mirror of the role pre-check rather than the real handlers, so a dropped `require_admin_or_above` wouldn't be caught — and the agent-card regression is a live example of a handler path the current suite misses. We'll add an end-to-end REST authority test that drives a real handler (one that would fail if the gate were removed), alongside the existing library-layer tests.

**Deferring the non-blocking items** to a follow-up: the stable-id alias discovery leak, the ban-of-sole-admin apply-path test, the mixed-version upgrade caveat, and the tombstone / error-string docs. One question rather than a unilateral change — adding a `withdrawn` field to the `GET /groups` response is an API surface change, so I'd rather check you want it here vs. as a fast-follow.

**On CI:** the red `Test Suite` is the `x0x_0041_prefer_newest_test` 500ms timing flake — same family as the already-excluded `x0x_0041_synthetic_kill_restart`, our branch touches none of that code, and it failed the re-run the same way (these flake under load rather than 50/50). Whenever convenient, folding `prefer_newest_test` into the same exclusion as its sibling would stop it gating unrelated PRs — happy to put that in a separate small PR. So a red on that specific test isn't a reason to hold the PR.

**On merging:** with the blocker and the three should-fix-now items addressed, please don't feel you need to wait on me — if it meets your bar, go ahead and merge and I'll pick up the deferred follow-ups (and anything else you flag) afterwards. Equally happy for you to hold it if you'd rather take the fixes and the follow-ups together. Either works.

Thanks again @dirvine! and for the thoroughness of the review. :)

JimCollinson:
Fixes are up as `bc9a904` — CI's green. Quick rundown of what landed and what's deferred.
 
**Done in this push:**
 
- **Agent-card invites (the blocker).** The card path now populates the base-state fields the join path needs — via a shared helper, so it can't drift from generate-invite — with a card-derived-invite join test (scoped to accepted-on-join; full card convergence stays Phase 2). The join-path check is unchanged. It also now skips exporting withdrawn/tombstoned groups, so a deleted group can't be re-advertised as a joinable invite.
- **TreeKEM TOCTOU.** The final withdrawn-check now runs while holding the `treekem_groups` write lock (bail + wipe if withdrawn); a deterministic race test guards it.
- **GUI.** A demote-to-member control (admin-gated; the last-admin 409 still applies), and the caller's own row no longer shows management actions.
**Rule 9** — agreed it's the sharpest note. The new card-invite test already drives the real handlers rather than a mirror; the dedicated end-to-end REST authority test (one that fails if an admin gate is dropped) I'd land as a fast-follow.
 
**Fast-follow (non-blocking, not in this PR):** that Rule 9 test, plus the stable-id alias discovery leak, the ban-of-sole-admin apply-path test, the mixed-version upgrade caveat, and the tombstone / error-string docs. None of these need to hold the merge — I'm happy to land them as a follow-on, and equally happy for you to pick any of them up directly if you'd rather they're in sooner. (Adding a `withdrawn` field to `GET /groups` is an API surface change — your call whether you even want it.)

dirvine (approved):
## Approving — all requested changes resolved + testnet-validated

My earlier `CHANGES_REQUESTED` is now fully addressed:

**Blocking + should-fix (resolved in `bc9a904`, each with a covering test):**
- 🔴 Agent-card invites unjoinable → `populate_invite_base_state_from_group_info` now applied in `get_agent_card`; +2 tests (accepted-on-join, withdrawn-not-exported)
- 🟠 TreeKEM install TOCTOU → re-check under the `treekem_groups` write lock; + race test
- 🟠 GUI admin-demote missing → "Demote to member" control + binding; + gui_smoke test
- 🟡 GUI self-row not excluded from Ban/Promote → `manage = canManage && !isSelf`; + gui_smoke test

**Polish (resolved in `9e2d57f`):**
- Greptile #1 — `assignable_from_name` gives a case-sensitivity hint for mis-cased valid roles
- Greptile #3 — dropped the redundant `clear_group_info_key_material` after `seal_withdrawal` (which already nulls `shared_secret`); the GSS-epoch/TreeKEM-plane sites stay direct assignment by intent (not terminal wipes)
- Rule 9 — added `membership_handlers_reject_non_admin_local_caller`: invokes the real remove/ban/role REST handlers with a non-admin caller and asserts 403, closing the gap that `tests/membership_authority.rs` (which re-implements the check) cannot catch
- `tests/CLAUDE.md` lists the 4 new ADR-0016 authority test files

**Validation:**
- `cargo fmt` + `clippy -D warnings` clean; affected + ADR-0016 group/authority suites green (69 tests)
- Codex (`codex-cli 0.133.0`) review of the polish commit: **SHIP**, no blocking issues
- Local 3-instance group+contacts dogfood: **33 pass / 0 fail**
- VPS testnet (isolated plane, 6 nodes): deploy clean; group create/invite/join, roster convergence on all nodes, group messaging, retained state-commits (#111), and the **TreeKEM applied path** all PASS cross-region. Residual non-passes were transport/DM-delivery to the APAC nodes (known PMTU/transit degradation), not authority logic.
- CI: green (the transient Test Suite / Multi-Agent / Cargo-Audit reds were environmental flakes — a different unrelated daemon/transport test each run, all passing locally — and cleared on re-run).

PR #327, opened 2026-08-15
feat(dm): DA slice 1 — capability advert + strict-send gate (ADR 0030 §2/§3)

dirvine:
First of four slices landing DM protocol v2 per **ADR 0030** (Accepted 2026-08-13). Sender side only; receiver durable path is slice 2.

## What lands (§ references = ADR 0030)

- **§2**: protocol constants with **ceiling `DM_PROTOCOL_VERSION = 2`** — above-ceiling envelopes dropped without ACK, with an observable trace event (ADR 0025); no v3/thread code active.
- **§3**: `v2_durable_gossip_ready` **iff durable history enabled** (the ADR's amendment over the campaign's v3-iff-history), advertised on `start_dm_inbox`/withdrawn on stop; `/agent/card` + A2A card export the live value. `CapabilityStore` never-lower rule on card import; signed-timestamp expiry with bounded skew.
- Targeted refresh protocol (request/response topics registered **Critical** so strict-send convergence isn't cooled behind the Bulk steady topic) + warm carrier on the steady topic; recipient-keyed single-flight to prevent advert storms.
- Strict gate: `require_durable_app_ack` (default **false**) → on cache miss, exactly one forced targeted refresh, bounded 3s wait, then `AckSemanticsUnavailable` → REST **409 `recipient_ack_semantics_unavailable`**. Never silent downgrade, never a hang.
- ACK-waiter binding: exact protocol match — a v1 ACK can never satisfy a v2 waiter.
- **§4 note (second commit)**: `require_gossip_ack` on `/direct/send` is *honored* on main today (the campaign's silent discard was the defect), so the hard-400 is deliberately deferred to slice 4 where product defaults change; current behavior is now documented in `api-reference.md` with the deprecation notice.

## Extraction provenance

Hunk-by-hunk from `wip/codex-durable-app-ack` (`b60b995`), never file blobs. The survey's "new files" (`dm_capability*.rs`) already existed on main — extracted as deltas. Excluded and verified absent: receiver durable path, raw-durable framing, `send_durable_raw_preferred`, outbox, product defaults, all v3/threaded code, `logical_id` derivation changes.

## Gates (verified independently on the pushed tree)

[...]

PR #328, opened 2026-08-15
feat(dm): DA slice 2 — receiver durable path (ADR 0030)

dirvine:
DA slice 2 of 4 per [ADR 0030](docs/adr/0030-dm-durable-application-ack-v2.md) (Accepted). Follows slice 1 (`d1bcc85`, #327). **Not for merge without cross-review.**

### 1. Receiver durable path (ADR 0030 §1)

`InboxPipeline::handle_payload_durable` runs a v2 payload in exactly the normative order:

1. **Per-logical-request lock** — `RecentDeliveryCache::delivery_lock`, serializing the primary-inbox and legacy-bus copies of one envelope. No free slot ⇒ ACK withheld; evicting an in-flight lock would permit duplicate dispatch.
2. **Replay-cache binding check** — an existing completion is re-ACKed under *its own* semantics, never re-dispatched. The binding is a domain-separated blake3 digest over `(protocol_version, accepted bytes)`, so the same request id carrying different content is refused rather than re-ACKed as the original.
3. **Durable-history lookup** — keyed on the schema v4 columns via the new `Store::find_by_logical_request`. This is the check that survives restart, where the memory-only replay cache does not.
4. **Dispatch.**
5. **`record_committed` awaited** — must return `Inserted | Duplicate`.
6. **ACK**, stamped with the semantics actually honoured.

**The ACK exists only because the commit succeeded.** Commit error, backpressured/closed writer, unavailable history, lost lock, or an unpublishable replay completion each *withhold* the ACK. No branch falls back to a v1 ACK: a withheld ACK costs the sender a timeout, whereas a downgraded one hands it a weaker receipt it cannot distinguish from the durable one it asked for (§2).

### 2. ACK version stamping

`publish_ack_for_protocol` replaces the hardcoded v1 stamp. v2 is passed **only** from the durable path, after `record_committed` returns. The v1 receive path is untouched: a v1 envelope takes the legacy path and gets a v1 ACK even with durable history enabled — test `v1_envelope_receives_v1_ack_even_with_durable_history_enabled`.

One deliberate refinement: a *refusal* (policy rejection, binding conflict) is stamped with the **requested** version rather than downgraded, because a refusal makes no durability claim and must reach the sender's exact-protocol waiter. Downgrading it would turn an actionable "no" into a timeout, which §2's drivers forbid. Only `Accepted` is constrained to the semantics actually honoured.

### 2b. `AckSemanticsUnavailable` as an ACK outcome (ADR 0030 §2)

ADR 0030 §2 names this outcome explicitly: *"A logical request completed under weaker semantics is answered `AckSemanticsUnavailable`, never re-ACKed as durable (`cached_ack_for_protocol`)."*

Adds `DmAckOutcome::AckSemanticsUnavailable { reason }`, used for both refusals the durable path can make — already completed under weaker semantics, and already bound to different content. `dm_send::ack_outcome_to_receipt` maps it to `DmError::AckSemanticsUnavailable`, the same typed error the slice-1 send-side capability gate raises.

This matters beyond naming: the alternative (`RejectedByPolicy`) is mapped by the sender to `DmError::RecipientRejected`, so a product UI would render a protocol condition as *"peer blocked you"* instead of *"retry / peer needs upgrade"* — misreporting the one thing the 409 contract exists to make actionable.

**Wire compatibility**, covered by two tests:
- The variant is **appended last**, so postcard discriminants for `Accepted` (0) and `RejectedByPolicy` (1) are unchanged — 0.37 peers keep decoding our ACKs. Asserted by `appending_the_new_ack_variant_preserves_legacy_discriminants`.
- A v1-only 0.37 peer can never *receive* it: it is produced only when the request asks for semantics above what completed, and a v1 sender never asks for more than v1. Asserted by `a_v1_request_is_never_answered_with_the_new_ack_variant`. Were this to regress, an 0.37 receiver would fail to decode the ACK and the send would degrade to a timeout.

### 3. ADR §7 typed-route dedupe gap — **DOCUMENTED AS A HANDLER OBLIGATION, NOT CLOSED**

I chose to **own the obligation**, not extend the lookup. Rationale — extending it is not merely more expensive, it would be *wrong*:

- Typed-prefix families (group ingest, KV deltas, exec audit, card import, voice signalling) classify **`Ephemeral`** in `history::classify` **by design**: their durable effect lives in their own store (group state chain, KV store, exec log), not in DM history. Committing them to DM history would duplicate the authoritative durable surface and grow history with plumbing traffic.
- Typed routes are handed off with a **non-blocking `try_send` that may drop** under backpressure. A durable ACK would claim a dispatch that never completed.

So a v2 envelope matching a typed route gets **no ACK** in this slice, and each typed handler carries the restart-spanning dedupe obligation via its own durable surface — the bootstrap outbox handler satisfies it with `Inserted | Duplicate` completion, exactly as the ADR anticipates. The typed-route completion signal (`DmTypedPayloadCompletion` on the campaign branch) that lets these frames earn a v2 ACK is **slice 3**. Locked in by `durable_typed_route_withholds_ack_pending_handler_completion`, so slice 3 has to change it deliberately.

### 4. Schema v4 columns gain writers

Inbound DM rows now populate `ingress_sender_agent` and `logical_request_id` (both paths, via the factored `inbound_dm_history_record`). New `Store::find_by_logical_request` reads them. No migration change — main is on the released v4 schema.

Also adds `HistoryHandle::record_committed` / `WriterHandle::record_committed` (a `WriteCommand::Commit` variant that flushes the pending batch first, preserving producer order) and `HistoryError::WriterBackpressured`. Unlike `record`, it never sheds silently.

### 5. Capability advert flip (ADR §3)

`start_dm_inbox` now advertises `v2_durable_gossip_ready(...)` **iff `history_handle.is_some()`**, else `v1_gossip_ready`. The slice-1 hold comment is replaced with the reason the flip is now correct. Protocol ceiling stays 2; no v3 anything.

### 6. Docs in the same PR (ADR §6)

- `CHANGELOG.md` — Unreleased entries for the durable path, v4 writers, `record_committed`, and a `### Changed` entry for the advert flip.
- `docs/design/dm-over-gossip.md` — normative receiver ordering, the §7 typed-route position, at-least-once-across-restart, the v2-iff-history advert rule, and delivery level 3 moved from "future" to "v2".
- `docs/api-reference.md` — note updated to slice 2. **No REST-visible shape change in this slice**; no new routes, so no manifest regeneration, no `api_coverage.rs` entry, no CLI subcommand needed.
- **CHANGELOG Notes tidy (reviewer nit from #327):** the interim-timeout caveat no longer reads as a general condition. It now states that a strict v2 send can time out *only against a peer advertising a false v2 capability*, and that this daemon cannot be that peer.

## Extraction method and omitted hunks

Extracted hunk-by-hunk from individual commit diffs on `wip/codex-durable-app-ack` (`b60b995`) with `git apply --reject`; no `git checkout --theirs`, no whole-file blobs. `src/history/writer.rs`, `src/error.rs` and `src/history/mod.rs` are touched on that branch **only** by `3d714db` (a durable-ACK commit, no contamination) and applied cleanly with zero rejects. Everything in `dm.rs`/`dm_inbox.rs` was hand-placed.

**Deliberately omitted:**

- **All campaign `src/history/store.rs` hunks.** They are the v2→v3→v4 migration work, which main already shipped in better form in v0.37.4 over the released FTS v2 schema. Dropping the campaign `store.rs` would have regressed the released schema — the explicit failure mode ADR 0030 "Neutral/Operational" warns about. The only store change here is the additive `find_by_logical_request` reader.
- **v3 threading.** `DmThreadMeta` does not exist on main and the ADR defers v3, so `DmDurableBindingDigest::accepted` and the logical-request binding drop the campaign's `thread_meta` parameter. Ceiling stays 2.
- **Hedged/multi-route ACK publisher** (`AckPublisherHandle`, `spawn_durable_ack_publisher`, `publish_durable_ack_routes`, `publish_raw_and_gossip_ack_routes`, `DURABLE_ACK_ROUTE_TIMEOUT` — from `bde4d28`, `fa538eb`, `ed04a23`, `21387d2`, `b60b995`). ACK *delivery* robustness, orthogonal to commit-before-ACK ordering.
- **Raw durable DM frame transport** (`RAW_DURABLE_DM_MAGIC`, `DmInboxIngress::handle_raw_frame`, `encode_raw_durable_frame`) — tied to the outbox slice.
- **`DmTypedPayloadCompletion`** — slice 3, per the §7 decision above.
- The campaign's `durable_history_logical_request` paginated 500-row scan that decodes every stored envelope in Rust, replaced by a single indexed-shape SQL lookup on the v4 columns. Same semantics (Missing/Exact/Conflict), far cheaper.

Legacy trust/origin/revocation tests in `dm_inbox.rs` were pinned to `DM_PROTOCOL_V1` via `craft_unsigned_payload_envelope`. They predate v2 and became v2 only incidentally when slice 1 raised `DM_PROTOCOL_VERSION`; they exercise gates that run before protocol version matters.

## Tests

New, all plain `#[tokio::test]` / `#[test]`:

- `durable_ack_is_withheld_when_history_commit_fails` — writer shut down mid-flight; no durable row, no replay completion.
- `durable_path_reports_withheld_ack_on_commit_failure` — asserts `Withheld("commit_failed")` **directly**. The durable path returns a `DurableAckDecision` so "no ACK when the commit fails" is a checked outcome, not an absence inferred from side effects.
- `durable_path_commits_before_acking_and_stamps_v2` — v2 stamp, exactly one row, both v4 columns populated.
- `durable_replay_rebinds_instead_of_committing_twice` — replay commits no second row.
- `durable_replay_with_different_bytes_is_refused` — binding conflict refused, not accepted.
- `v1_envelope_receives_v1_ack_even_with_durable_history_enabled` — mixed-version case.
- `v2_envelope_without_history_is_never_downgraded_to_v1` — withheld, no completion.
- `durable_typed_route_withholds_ack_pending_handler_completion` — §7 position.
- `cached_v1_completion_refuses_a_v2_request` — `cached_ack_for_protocol` returns `AckSemanticsUnavailable`, not `RejectedByPolicy`.
- `a_v1_request_is_never_answered_with_the_new_ack_variant` and `appending_the_new_ack_variant_preserves_legacy_discriminants` — wire safety for the appended ACK variant.
- `dm_capability_advert_tracks_durable_history_presence` (lib.rs) — asserts the advert in **both** directions, history on ⇒ 2 and off ⇒ 1.

## Gate results (real exit codes, no `| tail` on the gates)

| Gate | Exit |
|---|---|
| `cargo nextest run --workspace --all-features --no-fail-fast` | **0** — 2630 passed, 260 skipped, 0 failed |

No `.unwrap()` / `.expect()` / `panic!()` added to production code (all such matches in the diff are inside `#[cfg(test)]` / `#[tokio::test]`).

**Known flake, verified:** on one full run `direct_send_with_require_ack_round_trips_to_live_peer` failed with a client-side `reqwest ... TimedOut` on the localhost `/direct/send` call — a transport timeout, not an assertion. Re-run in isolation **3/3 passing**, and the clean `--no-fail-fast` full run above is 2630/2630. It is unreachable from this change's new code: that test sends with `require_durable_app_ack = false`, so the envelope is v1 and never enters the durable path or the new ACK variant. `suppressed_peer_inbound_redial_is_rejected` passed every run.

## Incidental finding (not fixed here)

`src/dm_inbox.rs` contains a literal NUL byte at line ~1947 — a test writes `b"x0x-exec-v1\0"` as a raw byte rather than the `\x00` escape. Harmless to `rustc`, but it makes `grep`/`rg` treat the whole file as binary and silently suppress matches past that offset, which is a real hazard when reviewing this file. Left alone to keep this PR surgical; worth a one-line follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,15 +6,38 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- **DM protocol v2 receiver durable path (ADR 0030 §1, slice 2 of 4).** A v2
+  envelope is answered in exactly this order: per-logical-request lock →
+  replay-cache binding check → durable-history lookup → dispatch →
+  `record_committed` awaited → ACK. A v2 ACK is emitted **only** after the
+  SQLite commit returns `Inserted | Duplicate`; every failure below that —
+  commit error, backpressured writer, unavailable history, lost delivery
+  lock, unpublishable replay completion — withholds the ACK rather than
+  degrading it. `Accepted` under v2 now means what ADR 0030 §1 says it
+  means: verified, durably committed, and dispatched.
+- **History schema v4 columns gain writers.** Inbound DM rows populate
+  `ingress_sender_agent` and `logical_request_id`; together they key the
+  durable-history lookup that lets a restarted receiver recognise a logical
+  request it already committed, via the new
+  `Store::find_by_logical_request`.
+- **`HistoryHandle::record_committed`.** A commit-receipt write for protocol
+  surfaces whose success promises durable history. Unlike the fire-and-forget
+  `record`, it never sheds silently: a full or closed writer queue returns
+  `HistoryError::WriterBackpressured` / `WriterClosed` to the caller.
 - **DM protocol v2 capability advertisement + strict-send gate (ADR 0030,
   slice 1 of 4).** `DM_PROTOCOL_VERSION` becomes the receive **ceiling** 2:
-  envelopes above it are dropped without an ACK. The production advert is
-  **held at `max_protocol_version = 1`** until the slice-2 receiver durable
-  path ships in the same binary — advertising v2 without it would be a
-  false capability and hang strict senders (PR #327 review; the ADR 0030
-  §3 v2-iff-history flip lands with slice 2). Agent cards export the same
+  envelopes above it are dropped without an ACK. Agent cards export the same
   runtime value, and card imports go through `CapabilityStore::insert_from_card`
   so a stale card can never lower a live signed advert.
+
+### Changed
+
+- **The v2 capability advert is now live and conditional (ADR 0030 §3).**
+  `start_dm_inbox` advertises `max_protocol_version = 2` **iff** a durable
+  history handle is present, else 1. The slice-1 hold is lifted because the
+  receiver path in this release is what makes the v2 claim true. Daemons
+  without history keep advertising v1, so a strict sender gets the typed 409
+  instead of clearing its gate against a receiver that could never commit.
 - **`DmSendConfig::require_durable_app_ack`** (default `false`). A strict
   send negotiates v2 on the wire and pins its ACK waiter to
   `(protocol_version, recipient, machine)`; a recipient without a current
@@ -32,12 +55,31 @@ All notable changes to this project will be documented in this file.
 ### Notes
 
 - No send surface sets `require_durable_app_ack` yet: REST, CLI, WS and the
-  library default stay v1-defaulted, so this slice is behaviour-neutral for
-  existing callers. Product defaults flip in ADR 0030 slice 4.
-- The receiver durable path is slice 2. Until it lands this build ACKs with
-  `protocol_version = 1` — the semantics it actually honours — so a strict
-  send that clears the gate times out rather than receiving a durable
-  receipt no one made.
+  library default stay v1-defaulted, so these slices remain behaviour-neutral
+  for existing callers. Product defaults flip in ADR 0030 slice 4.
+- **v1 senders are unaffected.** A v1 envelope takes the unchanged legacy
+  path and receives a v1 ACK even on a receiver with durable history enabled;
+  enabling history never silently upgrades the receipt an 0.37 peer is handed.
+- A strict v2 send can still end in a timeout, but now only against a peer
+  advertising a **false** v2 capability — one claiming the durable ceiling
+  without a receiver that can commit. This daemon cannot be that peer: it
+  advertises v2 only when history is enabled, and withholds the ACK rather
+  than downgrading if history becomes unavailable at runtime.
+- **Typed routes (ADR 0030 §7): the obligation is documented, not closed.**
+  Typed-prefix families (group ingest, KV deltas, exec audit, card import,
+  voice signalling) classify `Ephemeral` — their durable effect lives in
+  their own store, not in DM history — and are handed off with a
+  non-blocking `try_send` that may drop. A DM-history commit therefore
+  cannot honestly back their receipt, so a v2 envelope matching a typed
+  route gets **no** ACK in this slice, and each handler carries the
+  restart-spanning dedupe obligation via its own durable surface. The
+  typed-route completion signal that lets these frames earn a v2 ACK is
+  slice 3.
+- Durable v2 is **at-least-once across restart**: the replay cache is
+  memory-only, so a restart can re-dispatch an already-committed envelope.
+  The durable-history lookup keeps it to exactly one history row, and
+  applications needing restart-spanning exactly-once dedupe on
+  `(sender, request_id)`.
 
 ## [v0.37.4] - 2026-08-14
 
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -382,11 +382,16 @@ Notable codes:
 | 413 | `payload_too_large` | Payload exceeds the DM envelope cap. |
 | 504 | `timeout` | No application ACK within the retry budget. The DM may or may not have arrived. |
 
-**Note (as of ADR 0030 slice 1):** no REST field selects durable ACK
+**Note (as of ADR 0030 slice 2):** no REST field selects durable ACK
 semantics yet, so `recipient_ack_semantics_unavailable` is currently
 reachable only through the library (`DmSendConfig::require_durable_app_ack`).
 The product-tier default flips in ADR 0030 slice 4, at which point this
 becomes a routine response for not-yet-upgraded peers.
+
+Slice 2 landed the receiver durable path, so a library-level strict send now
+*succeeds* against a peer that advertises v2 — which it does only when that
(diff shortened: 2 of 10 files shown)


dirvine:
## Cross-review (ADR 0030 §1/§3/§7)

Walked `747f1be` against the slice-2 bar.

**Binding closed.** `start_dm_inbox` is `v2_durable_gossip_ready` iff `history_handle.is_some()`, else v1. Advert test covers both directions. A v2 `Accepted` ACK is only published after `record_committed` returns `Inserted | Duplicate`. Commit / lock / history-missing / replay-publish failures withhold. No v1 downgrade of a v2 request.

**ACK stamp.** v1 envelopes stay on the legacy path and get v1 ACKs even with history on (`v1_envelope_receives_v1_ack_even_with_durable_history_enabled`). **Refusal stamped with the requested version is the right call** — a refusal is not `Accepted` and must hit the exact-protocol waiter; downgrading it is the hang §2 forbids.

**§7.** Document-the-obligation, not extend-the-lookup. Typed v2 is withheld (and not dispatched) this slice; locked by `durable_typed_route_withholds_ack_pending_handler_completion`. Slice 3 owns the completion signal. Accept.

**store.rs.** Only additive `find_by_logical_request` on the released v4 columns. Campaign migration hunks omitted. Good.

**Tests** encode the why (commit-before-ACK, no v1 downgrade, mixed-version v1 receipt, binding conflict, advert flip). `DurableAckDecision` is the right seam.

Same-account cannot formal-approve. **Glance-approve. Merge on CI green.**

Non-blocking:
- After restart, `Exact` history still re-dispatches then `Duplicate`s — allowed (at-least-once-across-restart). A test that names that would be nice in slice 3.
- `find_by_logical_request` has no index yet. Fine at current scale; don't let it become a table scan later.
- NUL-byte in `dm_inbox.rs` still makes `rg` treat the file as binary. Slice 3 tidy as you said.

Issue #316, opened 2026-08-12
Flaky tests under full parallel load: direct_send_with_require_ack_round_trips_to_live_peer, suppressed_peer_inbound_redial_is_rejected


dirvine:
Observed during local validation of PR #315 (test-only branch on top of cd893b7), three consecutive `cargo nextest run --workspace --all-features` runs on the same tree:

1. Run 1: `x0x::peer_lifecycle_integration direct_send_with_require_ack_round_trips_to_live_peer` FAILED after 18.1s (passes in isolation in 8.3s). 2577/2578 otherwise green.
2. Run 2: `x0x tests::suppressed_peer_inbound_redial_is_rejected` FAILED fast (0.16s; passes in isolation in 2.2s, 4/4 attempts). Fail-fast stopped the run.
3. Run 3 (`--no-fail-fast`): 2578/2578 green.

Both are loopback QUIC tests and only fail under full-suite parallel contention — same class as #311 (a2a_binding_integration::unary_round_trip_over_dm). The failing runs shared no failing test, and the tree under test contains zero production changes, so this is load-dependent timing, not a regression.

Suggested directions: per-test retries are masking, so prefer either widening the ack/redial deadlines under load or moving the most contention-sensitive loopback tests to a serial test group (nextest `test-groups` with `max-threads = 1`), as consistent with ADR 0025's requirement that a pass means the observation actually completed.

Environment: macOS (Darwin 25.6.0), local dev machine, nextest default profile.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

dirvine:
Triage against `f3080c39b444b7ce44216c1811c472f3b4e938cb` (v0.41.3), 2026-09-05: **P2 — Mitigated; full-load closure evidence needed.**

nextest assigns the named tests to quic-localhost and reserves four threads; this matches the reported remediation. Those controls are present, but isolated passes cannot discharge a failure that only occurs under full-suite contention.

**Acceptance / next step:** Collect repeated complete-suite runs on current main with exact commands and no retries; retain failures and stage traces. Close only when the specified load-sensitive scenarios actually run.

This is a source-and-evidence triage; platform/soak tests are credited only where explicitly stated.

dirvine:
**Verification 2026-09-10 (main @ 0532f58)**

[...]

PR #329, opened 2026-08-15
fix(dm): restore AckSemanticsUnavailable ACK outcome dropped from #328

dirvine:
Follow-up to #328 (ADR 0030 slice 2). **This is not new scope — it is a commit that #328 was supposed to carry and lost.**

## What happened

I pushed slice 2 in two commits. `747f1be` (the receiver durable path) and `da87c05` (this change). #328 was squash-merged as `dc3e862` in the window between those two pushes, so main got the first commit and not the second. Verified: `git diff dc3e862 origin/feat/da-slice2-receiver-durable-path` is exactly this change — 107 insertions across 5 files — and `git branch --contains da87c05` lists only the old feature branch.

This is a clean cherry-pick of `da87c05` onto current `origin/main`, no conflicts, no edits.

## The defect currently on main

ADR 0030 §2 is explicit: *"A logical request completed under weaker semantics is answered `AckSemanticsUnavailable`, never re-ACKed as durable (`cached_ack_for_protocol`)."*

`DmAckOutcome` on main has only `Accepted` and `RejectedByPolicy`, so the durable path's refusals currently emit `RejectedByPolicy` — and `dm_send::ack_outcome_to_receipt` maps that to `DmError::RecipientRejected`.

The user-visible consequence: a v2 sender whose logical request already completed under v1 semantics, or is already bound to different content, is told **"the recipient rejected you"** when the truth is **"you cannot have a durable receipt for this request."** A product UI renders a protocol condition as a blocked peer — misreporting exactly the case the 409 contract exists to make actionable. It is also the wrong remediation: the caller should retry or surface "peer needs upgrade", not treat the relationship as refused.

## The fix

- Adds `DmAckOutcome::AckSemanticsUnavailable { reason }`, used for both refusals the durable path can make: already completed under weaker semantics, and already bound to different content.
- `dm_send` maps it to `DmError::AckSemanticsUnavailable` — the same typed error slice 1's send-side capability gate raises, so both directions of the same condition report identically.

## Wire compatibility

This variant rides on a signed, postcard-encoded ACK body, so it is proven rather than asserted:

- **Appended last**, so postcard discriminants for `Accepted` (0) and `RejectedByPolicy` (1) are unchanged and 0.37 peers keep decoding our ACKs — `appending_the_new_ack_variant_preserves_legacy_discriminants`.
[...]

PR #330, opened 2026-08-15
feat(dm): answer un-upgradable requests with AckSemanticsUnavailable (ADR 0030 §2)

dirvine:
Follow-up to #328 (ADR 0030 slice 2). The receiver durable path merged as `dc3e862` carrying `747f1be` only; this is the remaining delta, re-based onto current `origin/main` so it gets its own review rather than riding in on an approval that did not cover it.

Clean cherry-pick of `da87c05` onto `dc3e862` — no conflicts, no edits. Tree verified byte-identical to the original commit.

## Why this is a correctness fix, not a naming preference

ADR 0030 §2 is explicit:

> A logical request completed under weaker semantics is answered `AckSemanticsUnavailable`, never re-ACKed as durable (`cached_ack_for_protocol`).

`DmAckOutcome` on main has only `Accepted` and `RejectedByPolicy`, so the durable path's refusals currently emit `RejectedByPolicy` — and `dm_send::ack_outcome_to_receipt` maps that variant to `DmError::RecipientRejected`.

The user-visible consequence: a v2 sender whose logical request **already completed under v1 semantics**, or is **already bound to different content**, is told *"the recipient rejected you"* when the truth is *"you cannot have a durable receipt for this request."*

That is wrong in two ways that compound:

1. **Wrong message.** A product UI renders a protocol condition as a blocked peer — misreporting precisely the case ADR 0030's 409 contract exists to make actionable.
2. **Wrong remediation.** `RecipientRejected` tells the caller the relationship is refused, so the correct responses (retry, or surface "peer needs upgrade") look inapplicable. The condition is transient and addressable; the error says it is neither.

## The fix

- Adds `DmAckOutcome::AckSemanticsUnavailable { reason }`, used for both refusals the durable path can make: already completed under weaker semantics, and already bound to different content.
- `dm_send` maps it to `DmError::AckSemanticsUnavailable` — the same typed error slice 1's send-side capability gate already raises — so both directions of the same condition report identically instead of diverging by which side noticed first.

## Wire safety

This variant rides on a **signed, postcard-encoded ACK body**, so it is proven by test rather than asserted in review:

**1. Legacy discriminants are preserved** — `appending_the_new_ack_variant_preserves_legacy_discriminants`.
Postcard encodes enum variants by declaration index. The new variant is **appended last**, so `Accepted` stays 0 and `RejectedByPolicy` stays 1, and every 0.37 peer keeps decoding our existing ACKs unchanged. The test pins all three discriminants, so a future reorder that would silently break the fleet fails here instead.

**2. A v1-only peer can never receive the new variant** — `a_v1_request_is_never_answered_with_the_new_ack_variant`.
`cached_ack_for_protocol` returns it only when the incoming request asks for semantics **above** what the cached completion made. A 0.37 peer only ever sends v1 envelopes, so it asks for at most v1, and any cached completion is at least v1 — the branch is unreachable for it. The test sweeps both cached protocol versions against both pre-existing outcomes to hold that closed.

This second property is the load-bearing one: were it to regress, an 0.37 receiver would fail to postcard-decode the ACK, drop it, and the sender would degrade to a **silent timeout** — the exact black-hole failure mode ADR 0030 §2 forbids. The test guards a real regression path, not the shape of the code.

`cached_v1_completion_refuses_a_v2_request` is updated to assert the ADR-named outcome, with a comment recording why `RejectedByPolicy` is wrong here so the substitution is not quietly reintroduced.

## Gates — re-run on this branch against the new base (`dc3e862`), not carried over

| Gate | Exit |
|---|---|
| `cargo nextest run --workspace --all-features --no-fail-fast` | **0** — 2630 passed, 260 skipped, 0 failed |

No production-code `.unwrap()` / `.expect()` / `panic!()` added; all such matches in the diff are inside `#[cfg(test)]` / `#[tokio::test]`.

## Flake encountered and cleared — plus a config gap worth a separate fix

One full-suite run failed `gossip_plane_isolation::gated_and_open_pair_converges_after_legacy_grace`. **This is not one of the two documented flakes**, so I did not assume:

- It failed in **2.4 s** at `tests/gossip_plane_isolation.rs:43`, on the very first `connect_addr` (`All connection strategies failed: DirectIPv4: Timeout; DirectIPv6: Happy Eyeballs timed out`) — transport setup, before any application logic.
- That file contains **zero** DM usage (`send_direct`, `DmAckOutcome`, `require_durable`, `/direct/send` all absent), so this change cannot reach it.
- Isolated re-runs: **4/4 passing** at ~12.4 s, matching its normal runtime. The clean full run recorded above is 2630/2630.

**Separate finding, deliberately not fixed here:** `.config/nextest.toml` serialises localhost-QUIC-endpoint tests via the `quic-localhost` group precisely because "ant-quic's dual-stack socket handling causes intermittent connection failures ... when many endpoints share the loopback interface." `gossip_plane_isolation` creates exactly that topology but is **not** in the group or in the `threads-required` reservation added for issue #316. That looks like a coverage gap in the serial-group filters. I left it alone: editing shared test-scheduling config would be unrelated scope in a cherry-pick, and it can shift timing for other suites, so it deserves its own change and review.

<h3>Greptile Summary</h3>

This PR distinguishes unavailable durable-ACK semantics from recipient policy rejection and propagates the condition through the sender as the existing typed error.
- Appends `AckSemanticsUnavailable` without changing legacy postcard discriminants.
- Emits the new outcome for weaker cached completions and conflicting durable bindings.
- Maps the outcome to `DmError::AckSemanticsUnavailable` and documents the protocol behavior.
- Adds receiver-side protocol and discriminant compatibility tests.

The PR appears safe to merge, with only a non-blocking test-coverage gap around the new sender-side error conversion.

The receiver paths are gated to durable-protocol requests, legacy discriminants remain stable, and current consumers handle the new outcome correctly; only the key ACK-to-error mapping lacks direct regression coverage.

**Files Needing Attention:** src/dm_send.rs

| Filename | Overview |
|----------|----------|
| src/dm.rs | Appends the new ACK outcome while preserving the two legacy postcard discriminants. |
| src/dm_inbox.rs | Uses the new outcome for weaker cached semantics and conflicting durable bindings, with compatibility tests. |
| src/dm_send.rs | Correctly maps the new outcome to the typed semantics error, but lacks a focused regression test for that mapping. |
| docs/design/dm-over-gossip.md | Documents why durable-semantics refusals are distinct from trust-policy rejection. |
| CHANGELOG.md | Records the new outcome, sender behavior, and wire-compatibility constraints. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
  participant Sender
  participant Inbox
  participant Cache as Durable history/cache
  Sender->>Inbox: v2 durable DM request
  Inbox->>Cache: Check logical request and binding
  alt Existing completion satisfies request
    Inbox-->>Sender: Accepted / cached outcome
  else Weaker completion or conflicting content
    Inbox-->>Sender: AckSemanticsUnavailable
    Sender->>Sender: Map to DmError::AckSemanticsUnavailable
  end
```

`````markdown
### Issue 1
src/dm_send.rs:501-503
**Test the new ACK mapping**

The focused conversion tests cover `Accepted` and `RejectedByPolicy` but not this new `AckSemanticsUnavailable` arm. Add a case that pins the typed error so a future mapping regression cannot restore the incorrect recipient-rejection response unnoticed.

---

For each issue above, determine whether it is valid and should be fixed. If so, fix it directly.
`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(dm): answer un-upgradable requests ..."](https://github.com/saorsa-labs/x0x/commit/269f026795685e9fd07433bc073b5aab363172e1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53607022)</sub>

> Greptile also left **1 inline comment** on this PR.

**Context used:**

- Knowledge Base — [Direct Messaging: Send, Capability Gate, Inbox, and Forward](https://app.greptile.com/saorsa-labs/-/custom-context/knowledge-base/saorsa-labs/x0x/-/docs/dm-messaging.md)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,17 @@ All notable changes to this project will be documented in this file.
   lock, unpublishable replay completion — withholds the ACK rather than
   degrading it. `Accepted` under v2 now means what ADR 0030 §1 says it
   means: verified, durably committed, and dispatched.
+- **`DmAckOutcome::AckSemanticsUnavailable { reason }`** (ADR 0030 §2). The
+  receiver's answer when it cannot issue the durable receipt a request asked
+  for — the logical request already completed under weaker semantics, or is
+  already bound to different content. It maps to
+  `DmError::AckSemanticsUnavailable`, the same typed error the send-side
+  capability gate raises, rather than `RecipientRejected`: nothing about the
+  trust relationship failed, so callers surface "retry / peer needs upgrade",
+  not "peer blocked you". Appended last in the enum, so postcard variant
+  indices for `Accepted` and `RejectedByPolicy` are unchanged and 0.37 peers
+  keep decoding them; a v1-only sender can never be sent the new variant,
+  because it never asks for semantics above v1.
 - **History schema v4 columns gain writers.** Inbound DM rows populate
   `ingress_sender_agent` and `logical_request_id`; together they key the
   durable-history lookup that lets a restarted receiver recognise a logical
--- a/docs/design/dm-over-gossip.md
+++ b/docs/design/dm-over-gossip.md
@@ -235,7 +235,14 @@ order, per ADR 0030 §1:
    under *its own* semantics, never re-dispatched. The binding is a
    domain-separated digest over `(protocol_version, accepted bytes)`, so the
    same request id carrying different content is refused rather than
-   re-ACKed as the original delivery.
+   re-ACKed as the original delivery. Both refusals — "already completed
+   under v1 semantics" and "already bound to different content" — are
+   answered `DmAckOutcome::AckSemanticsUnavailable`, never
+   `RejectedByPolicy`: no trust decision failed, and the sender maps the
+   latter to `RecipientRejected`, which would misreport a protocol condition
+   as a blocked peer. On the sender it becomes
+   `DmError::AckSemanticsUnavailable` — the same typed error the send-side
+   capability gate raises.
 3. **Durable-history lookup** — keyed on the schema v4 columns
    `(ingress_sender_agent, logical_request_id)` via
    `Store::find_by_logical_request`. This is the check that survives restart,
--- a/src/dm.rs
+++ b/src/dm.rs
@@ -269,6 +269,20 @@ pub enum DmAckOutcome {
     /// that prefer silent rejection can set `trust.silent_reject = true`
     /// and skip emitting this ACK entirely.
     RejectedByPolicy { reason: String },
+    /// The recipient cannot issue the durable receipt this request asked for:
+    /// the logical request already completed under weaker semantics, or is
+    /// already bound to different content (ADR 0030 §2). Distinct from
+    /// [`Self::RejectedByPolicy`] — nothing about the trust relationship
+    /// failed, so the caller should surface "retry / peer needs upgrade"
+    /// rather than "peer rejected you".
+    ///
+    /// **Wire compatibility:** appended last, so postcard's variant indices
+    /// for `Accepted` (0) and `RejectedByPolicy` (1) are unchanged and an
+    /// 0.37 peer still decodes those. A 0.37 peer can never *receive* this
+    /// variant: it is only produced when the request being answered asked for
+    /// semantics above what completed, and a v1-only sender never asks for
+    /// more than v1 (see `cached_ack_for_protocol`).
+    AckSemanticsUnavailable { reason: String },
 }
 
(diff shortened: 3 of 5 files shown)


dirvine:
Independently verified: head is `269f026795685e9fd07433bc073b5aab363172e1`, same SHA I approved on #329, diff empty. Approve transfers. Merge on CI green.

PR #331, opened 2026-08-15
feat(dm): DA slice 3 part 1 — typed-route completion signal + carried fixes (ADR 0030 §7)

dirvine:
ADR 0030 slice 3, **part 1 of 2**. Everything the outbox depends on, plus the carried tidies. The outbox module itself is deliberately **not** in this PR — see "Why this is split" below.

Do not merge; cross-review as usual.

### 1. Typed-route completion signal (ADR 0030 §7) — the obligation upgrade
Slice 2 withheld every v2 ACK on a typed route, because typed-prefix families classify `Ephemeral` and no DM-history commit could honestly back their receipt. This replaces the blanket refusal with a completion contract:

```rust
pub enum DmTypedPayloadCompletion { Inserted, Duplicate }
pub type DmTypedPayloadCompletionResult = Result<DmTypedPayloadCompletion, String>;
```

`DmTypedPayload` gains `request_id` (so handlers dedupe on `(sender, request_id)` across restart) and a `completion` oneshot the inbox awaits. For a durable typed payload this replaces steps 3–5 of the receiver ordering — `Inserted | Duplicate` is the same "exactly one durable record" proof `record_committed` gives the generic path.

**`DmTypedPayload` loses `Clone`**, since a oneshot sender has no meaningful copy and cloning would silently strand one caller's receipt. Audited: nothing cloned the payload; four construction sites updated.

**Per-route policy, stated (not accidental):** routes opt in via `with_durable_typed_payload_route`. A route that has not opted in still gets no v2 ACK, under its own distinct trace stage — this daemon will not certify durability for a handler that never promised it. Optimistically resolving before the store commit would convert a durable receipt into a lie; the honest failure is to drop the sender.

**All four non-completions withhold:** channel unavailable, handler error, dropped sender (so an early return on a handler's own error path fails safe), timeout.

**Deliberate deviation from the campaign draft.** It awaited completion for the envelope's remaining lifetime — up to 120 s — inline on the serial inbox loop, while the sender's per-attempt budget is 30 s (`DM_TIMEOUT_MAX_MS`). That lets one slow handler pin the entire DM pipeline for an ACK nobody is still waiting on. Bounded here at **20 s**: a receiver-side liveness bound, not a correctness one — a timeout withholds and the sender retries.

The slice-2 lock `durable_typed_route_withholds_ack_pending_handler_completion` is replaced, which is exactly what it existed for.

### 2. `find_by_logical_request` index (reviewer non-blocking from #328)
Partial index on `(ingress_sender_agent, logical_request_id) WHERE logical_request_id IS NOT NULL` — only rows written by the receiver durable path populate the v4 columns, so indexing group rows and every pre-v4 row would cost space for entries the lookup can never match. Mirrors the existing `idx_replace` precedent.

**Created idempotently at open, not as a v4→v5 migration.** An index changes no column and no row meaning, so an older binary reads and writes the database unchanged. A version bump would instead make `migrate()` reject the db as "newer than this binary" — so a rollback to v0.37.4 would leave every upgraded user with a `history.db` that refuses to open. Wrong price for a performance-only change; data-model changes still go through the versioned chain.

Tests assert the **query plan** uses the index (an index SQLite declines to use is indistinguishable from none at runtime) and that a db already stamped v4 gains it on open without a version bump.

### 3. Restart re-dispatch documented (ADR-allowed, previously unstated)
Verified against the code first: `Missing` and `Exact` share a fall-through, so an `Exact` hit after restart re-dispatches, re-commits to `Duplicate`, and re-ACKs v2. Documented why that's correct — skipping dispatch on `Exact` would silently drop any message whose crash landed between the commit and the application finishing with it, the exact window this protocol closes. Guarantee stated as **exactly one history row, at least one dispatch**.

### 4. `gossip_plane_isolation` flake — the briefed fix would have been a no-op
`binary(gossip_plane_isolation)` was **already** in the `quic-localhost` group; adding it again would have changed nothing while carrying a fix-shaped commit message. The real gap is that its override sets `test-group` but not `threads-required` — the issue-#316 signature already documented in that file: a group bounds contention *within* the group, but the trigger is ambient load from the ~2600 ungrouped tests starving the dial's timing window. Our failure matched exactly: dead at the first `connect_addr` in 2.4 s against a ~12.4 s normal runtime, 4/4 green isolated.

Nextest applies the **first** matching override (confirmed against the nextest docs), so appending the setting to the existing later block would have silently done nothing — the same failure mode one layer down. A narrower binary-only override placed ahead of it fixes the flake without changing the three slice-1 tests that share the broader block. `nextest show-config test-groups` confirms the binding.

### 5. Raw NUL bytes escaped, with a tripwire — **scope corrected from my earlier report**
`src/dm_inbox.rs` carried four raw NUL bytes where `\x00` escapes were meant.

**Correction, stated plainly:** I earlier reported this as the Panic Scanner "failing open". **That was wrong.** My shell's `grep` is a Claude Code wrapper around `ugrep` that always passes `-I`; the scanner runs under bash with `/usr/bin/grep`, which sees all 61 matches in that file — GNU/BSD grep classify from an early read buffer and these NULs sat at byte ~114788 of 117295. The gate was never blind, and no past approval needs recalibrating on that basis.

What was real: **rg/ugrep-based code search over that file returned silent false negatives**, which is how it was found. Developer-tooling correctness, not CI integrity.

The tripwire now tests for **raw NUL bytes directly** rather than "some grep thinks this is binary" — deterministic across tools and NUL positions, and it would have caught the original. Verified by injecting a NUL (gate fails) and restoring (gate clears). `-a` stays on the scan as defence in depth, since a NUL landing early genuinely would trip real grep.

## Why this is split

The outbox depends on the completion signal — on main, `handle_payload_durable` returned before dispatch for typed routes, so a strict v2 bootstrap DM never reached the handler at all; an outbox without the signal livelocks at 60 s backoff until cap eviction. So the signal had to land first regardless.

The outbox port itself is ~430 lines of function bodies plus adaptation across 4 files, and the adaptation is the hard part rather than the copying: the campaign version calls `send_direct_with_config_and_thread(..., Option<DmThreadMeta>, Option<DmLogicalId>)` and a `DmLogicalId` type, **neither of which exists on main**, so it needs a new logical-id-carrying send entry point designed alongside it. Combined with fail-closed sidecar startup semantics (a malformed outbox file aborts daemon startup) and a hard size guardrail on `named_groups.rs`, that is work I would rather land as its own reviewable change than rush onto the end of this one. A half-ported outbox that aborts startup is worse than no outbox.

Everything in this PR is complete, tested, and independently useful; nothing here is a partial step that only makes sense once the outbox lands.

## Gates (real exit codes)

| Gate | Exit |
|---|---|
| `cargo nextest run --workspace --all-features --no-fail-fast` | **0** — 2634 passed, 260 skipped, 0 failed |
| `scripts/check-panics.sh` | **0** |

No flakes on the full run, including `gossip_plane_isolation` under its new reservation. No production `.unwrap()`/`.expect()`/`panic!()` added.

<h3>Greptile Summary</h3>

The PR introduces handler-backed completion for durable typed DMs, adds the logical-request history index, strengthens the raw-NUL source gate, and adjusts nextest resource reservations.

- Adds request IDs and one-shot durable completion to typed payload delivery.
- Adds an idempotently created partial SQLite index for durable-request lookup.
- Documents restart redispatch semantics and updates associated tests.
- Reserves scheduler slots for `gossip_plane_isolation` and checks Rust sources for raw NUL bytes.

The durable typed-route API needs correction before merging because valid history-disabled configurations withhold every v2 request before dispatching it.

The new handler-owned durability path is placed after an unconditional generic-history availability check, making the advertised replacement path unusable for library embedders that retain the documented history-disabled default.

**Files Needing Attention:** src/dm_inbox.rs

| Filename | Overview |
|----------|----------|
| src/dm_inbox.rs | Adds the typed-handler completion protocol, but an early generic-history gate prevents durable typed routes from operating when history is disabled. |
| src/history/store.rs | Adds an idempotent partial index after migration; supported schema states contain the indexed columns and exclusive-open behavior remains intact. |
| scripts/check-panics.sh | Adds deterministic raw-NUL detection and forces grep to scan binary-classified inputs as text. |
| .config/nextest.toml | Adds a first-match binary-specific scheduler reservation while preserving the broader serialization override. |
| docs/design/dm-over-gossip.md | Documents typed-route completion and restart redispatch semantics consistently with the new flow, aside from the implementation's residual history dependency. |

</details>

<h3>Flowchart</h3>

```mermaid
%%{init: {'theme': 'neutral'}}%%
flowchart TD
  V2[v2 typed DM] --> H{History enabled?}
  H -- No --> W[Withhold without dispatch]
  H -- Yes --> R{Durable typed route?}
  R -- No --> G[Generic history path]
  R -- Yes --> D[Dispatch to typed handler]
  D --> C{Completion}
  C -- Inserted or Duplicate --> A[Cache completion and ACK]
  C -- Error, drop, or timeout --> X[Withhold and retry]
```

<details open><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `src/dm_inbox.rs`, line 1020-1029 ([link](https://github.com/saorsa-labs/x0x/blob/77e6f0c41d5bca3cb25f99b60d31ecae4dc1bb2b/src/dm_inbox.rs#L1020-L1029)) 

   <a href="#">P1</a> **History gate blocks typed durability**

   When a v2 payload matches a durable typed route while DM history is disabled, this early return withholds the request before route classification or handler dispatch, causing every attempt to retry or fail even though the handler’s own store is the durable surface.

   **Knowledge Base Used:**
   - [Direct Messaging: Send, Capability Gate, Inbox, and Forward](https://app.greptile.com/saorsa-labs/-/custom-context/knowledge-base/saorsa-labs/x0x/-/docs/dm-messaging.md)
   - [Durable local history (ADR-0023)](https://app.greptile.com/saorsa-labs/-/custom-context/knowledge-base/saorsa-labs/x0x/-/docs/history.md)
   
   <details><summary>Prompt To Fix With AI</summary>
   
   `````markdown
   This is a comment left during a code review.
   Path: src/dm_inbox.rs
   Line: 1020-1029
   
   Comment:
   **History gate blocks typed durability**
   
   When a v2 payload matches a durable typed route while DM history is disabled, this early return withholds the request before route classification or handler dispatch, causing every attempt to retry or fail even though the handler’s own store is the durable surface.
   
   **Knowledge Base Used:**
   - [Direct Messaging: Send, Capability Gate, Inbox, and Forward](https://app.greptile.com/saorsa-labs/-/custom-context/knowledge-base/saorsa-labs/x0x/-/docs/dm-messaging.md)
   - [Durable local history (ADR-0023)](https://app.greptile.com/saorsa-labs/-/custom-context/knowledge-base/saorsa-labs/x0x/-/docs/history.md)
   
   ---
   
   For each issue above, determine whether it is valid and should be fixed. If so, fix it directly.
   `````
   </details>
</details>

`````markdown
### Issue 1
src/dm_inbox.rs:1020-1029
**History gate blocks typed durability**

When a v2 payload matches a durable typed route while DM history is disabled, this early return withholds the request before route classification or handler dispatch, causing every attempt to retry or fail even though the handler’s own store is the durable surface.

---

For each issue above, determine whether it is valid and should be fixed. If so, fix it directly.
`````

</details>

<sub>Reviews (1): Last reviewed commit: ["docs(dm): record the typed-route complet..."](https://github.com/saorsa-labs/x0x/commit/77e6f0c41d5bca3cb25f99b60d31ecae4dc1bb2b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53614179)</sub>

<details><summary><h4>Context used (3)</h4></summary>

- Knowledge Base — [Direct Messaging: Send, Capability Gate, Inbox, and Forward](https://app.greptile.com/saorsa-labs/-/custom-context/knowledge-base/saorsa-labs/x0x/-/docs/dm-messaging.md)
- Knowledge Base — [Durable local history (ADR-0023)](https://app.greptile.com/saorsa-labs/-/custom-context/knowledge-base/saorsa-labs/x0x/-/docs/history.md)
- Knowledge Base — [Required Test Gates: Observation Completeness](https://app.greptile.com/saorsa-labs/-/custom-context/knowledge-base/saorsa-labs/x0x/-/docs/test-gate-policy.md)
</details>

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -41,11 +41,31 @@ filter = "test(direct_send_with_require_ack_round_trips_to_live_peer) | test(=te
 test-group = "quic-localhost"
 threads-required = 4
 
+# ADR 0030 slice 3: gossip_plane_isolation was ALREADY in the serial group
+# (see the block below) and still failed a full-parallel run — at the very
+# first `connect_addr`, in 2.4 s versus its normal ~12.4 s, with every dial
+# strategy timing out; 4/4 green in isolation. That is the issue-#316
+# signature exactly: group placement bounds group-internal contention, but
+# the trigger here is ambient load from the ~2600 ungrouped tests starving
+# the dial's timing window. So it needs the same remedy the two #316 tests
+# needed — a scheduler-slot reservation, not merely a group.
+#
+# This block MUST stay ahead of the broader one below: nextest applies the
+# FIRST matching override, so appending these settings after that block
+# would silently do nothing. Keeping the binary filtered separately also
+# leaves the three slice-1 tests below on their existing settings rather
+# than handing them reservations they have not been shown to need.
+[[profile.default.overrides]]
+filter = "binary(gossip_plane_isolation)"
+test-group = "quic-localhost"
+threads-required = 4
+
 # The slice-1 regression tests each stand up multiple loopback QUIC
 # endpoints; unserialized they add enough contention to break the
 # gossip_plane_isolation stay-down window (2-of-2 full-run repro, green
 # with these excluded, green on the v0.37.1 baseline). The plane-isolation
-# binary itself is multi-endpoint loopback and belongs in the group too.
+# binary itself is multi-endpoint loopback and belongs in the group too
+# (now via the more specific override above, which matches it first).
 [[profile.default.overrides]]
 filter = "test(discovered_but_disconnected_peer_is_dialed_by_same_send) | test(identity_epoch_converges_with_linear_application_publishes) | test(non_treekem_self_leave_directly_delivers_member_removed_to_retained_owner) | binary(gossip_plane_isolation)"
 test-group = "quic-localhost"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,23 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- **Typed-route durable completion signal (ADR 0030 §7, slice 3).** Slice 2
+  withheld every v2 ACK on a typed route because typed-prefix families
+  classify `Ephemeral` and cannot be backed by a DM-history commit. A handler
+  can now report `DmTypedPayloadCompletion::{Inserted, Duplicate}` on a
+  oneshot carried in `DmTypedPayload`, and that signal releases the ACK.
+  `DmTypedPayload` also carries the envelope `request_id` so handlers can
+  dedupe on `(sender, request_id)` across restart, and is no longer `Clone`
+  (a oneshot sender has no meaningful copy). Routes opt in via
+  `DmInboxConfig::with_durable_typed_payload_route`; routes that do not opt in
+  still get no v2 ACK, by stated policy. Every non-completion — channel
+  unavailable, handler error, dropped sender, timeout — withholds the ACK.
+- **`Store::find_by_logical_request` is now indexed.** Partial index on
+  `(ingress_sender_agent, logical_request_id)` for rows where
+  `logical_request_id IS NOT NULL`, created idempotently at open rather than
+  via a schema-version bump, so a rollback to v0.37.4 can still open the
+  database.
+
 - **DM protocol v2 receiver durable path (ADR 0030 §1, slice 2 of 4).** A v2
   envelope is answered in exactly this order: per-logical-request lock →
   replay-cache binding check → durable-history lookup → dispatch →
--- a/docs/design/dm-over-gossip.md
+++ b/docs/design/dm-over-gossip.md
@@ -284,6 +284,25 @@ already-committed envelope. The step-3 lookup holds it to exactly one
 history row. Applications needing restart-spanning exactly-once must dedupe
 on `(sender, request_id)`.
 
+Concretely, on a `DurableLogicalRequestLookup::Exact` after restart — the
+row is already committed and binds the same bytes — the receiver **still
+dispatches to the application again**, then re-commits (the store returns
+`Duplicate`, satisfying `exact_durable_history_outcome`) and re-ACKs v2.
+This is deliberate and ADR-0030-allowed, not an oversight:
+
+- The alternative, skipping dispatch on `Exact`, would silently drop the
+  message whenever the crash happened *between* the history commit and the
+  application actually finishing with it — the exact window this protocol
+  exists to close. A duplicate the application can dedupe is strictly safer
+  than a loss it cannot detect.
+- The receipt stays honest either way: the sender's `Accepted` means
+  committed **and** dispatched, and both are true on the replay.
+
+So the guarantee is precisely: **exactly one history row, at least one
+dispatch.** Anything an application does on receipt that is not idempotent
+must dedupe on `(sender, request_id)` — the same obligation §1 places on
+applications requiring restart-spanning exactly-once.
+
 ## Capability negotiation
 
 ### Advertisement
--- a/docs/design/dm-over-gossip.md
+++ b/docs/design/dm-over-gossip.md
@@ -263,19 +263,43 @@ distinguish from the durable receipt it asked for (ADR 0030 §2).
 
 #### Typed routes (ADR 0030 §7)
 
-The typed-route restart-spanning dedupe gap is **documented as a handler
-obligation, not closed in the inbox.** Typed-prefix families (group ingest,
-KV deltas, exec audit, card import, voice signalling) classify `Ephemeral`:
-their durable effect lives in their own store, not in DM history, so a
-DM-history commit could not honestly back their receipt — and they are
-handed off with a non-blocking `try_send` that may drop under backpressure,
-so a durable ACK would claim a dispatch that never completed.
+Typed-prefix families (group ingest, KV deltas, exec audit, card import,
+voice signalling) classify `Ephemeral`: their durable effect lives in their
+own store, not in DM history, so a DM-history commit could never honestly
+back their receipt. Slice 2 therefore withheld every v2 ACK on a typed route.
 
-A v2 envelope matching a typed route therefore receives **no** ACK, and each
-typed handler carries the dedupe obligation via its own durable surface (the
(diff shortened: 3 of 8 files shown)


dirvine:
## Cross-review (slice 3a / ADR 0030 §7)

Walked `77e6f0c` (6 commits, 590/46) against the slice-3 bar that applies to this half.

**Split is right.** Outbox needs this signal first (without it a v2 typed bootstrap never reaches the handler). Fail-closed sidecar + `named_groups.rs` size belongs in 3b. `named_groups.rs` stays **28274**.

**§7.** Slice-2 lock replaced on purpose. Opt-in via `with_durable_typed_payload_route`; non-opt-in withholds (`typed_route_not_durable`). `Inserted | Duplicate` release the ACK; handler error, dropped sender, channel full, and 20s timeout withhold. `Clone` removed on `DmTypedPayload` — correct. Tests name the why.

**Greptile's history-first note is not a 3a blocker.** `handle_payload_durable` still returns `history_unavailable` before typed dispatch. That matches the live advert rule (v2 iff `history_handle`). A history-off library advertises v1, so honest v2 traffic never arrives. **3b must not advertise v2 without history.** If you later want typed-only v2, move the history check after the typed branch *and* change the advert rule in the same PR. Do not do that here.

**Production routes still use `with_typed_payload_route`.** After merge, live typed v2 still withholds. That's fine for 3a (infrastructure). **3b must opt the outbox in** or the outbox still livelocks.

**20s HOL on the serial inbox loop** — accepted with eyes open. Better than the draft's 120s. Don't grow it in 3b.

**Index without a schema bump** — right rollback call. Query-plan test is the one that matters. Restart-Exact re-dispatch docs match the code. nextest override is first-match and ahead of the broader block. NUL tripwire tests the defect, not a grep heuristic; `-a` is defence in depth. Scope correction is in the PR body. Good.

**Approve 3a. Merge on CI green.** Same-account, no formal approve. 3b still carries the full §5 bar (`LegacyV1Sent` is never completion; outbox own module; 1024/60s; cleared only by frontier-matching v2 ACK).

PR #332, opened 2026-08-15
feat(groups): DA slice 3b — durable bootstrap outbox module (ADR 0030 §5)

dirvine:
## Scope

DA slice 3b: the durable signed-public bootstrap outbox, extracted into its own
module as ADR 0030 §5 requires. Ported hunk-by-hunk from the single campaign
commit `4880a9f` on `wip/codex-durable-app-ack`; none of the retained-redelivery
family on that branch is included.

Adding a member to a SignedPublic group used to direct-send the roster snapshot
fire-and-forget. An offline recipient never got it and nothing on the authority
remembered the debt. The send is now an obligation:

- keyed by `(recipient, group, frontier, payload-digest)` as one blake3 binding
  digest, which is also the DM logical request id;
- persisted to `public_group_bootstrap_outbox.json`, written in the same step as
  the roster commit (outbox first, rolled back if the roster write fails);
- capped at 1024 entries; retried on `1s << min(attempts, 6)` clamped to 60 s;
- **fail-closed on load** — version mismatch, over-cap, duplicate key, or a
  payload that contradicts its own frontier aborts daemon startup;
- discharged **only** by a frontier-matching v2 application ACK. `LegacyV1Sent`
  reschedules; a failed send reschedules; neither is completion.

The route is registered with `with_durable_typed_payload_route` and its handler
resolves `typed.completion` with `Inserted | Duplicate` only after the consent
gate and a directory-durable install. No v2 advert changes; ceiling stays 2;
nothing v3; no product-default flips.

New module: `src/server/routes/public_group_bootstrap_outbox.rs` (1,544 lines,
of which ~480 are its unit tests). It owns the bootstrap protocol end to end —
the typed wire format, the obligation and its persistence, the retry engine, and
both halves of the receiving side (`admit_public_group_bootstrap` and the typed
handler that resolves the completion). `named_groups.rs` keeps group CRUD.

### Completion ordering (reviewer requirement 1)

A v2 ACK certifies durability, and the sender **deletes its obligation** on the
strength of it. So every completion here is gated on *directory-durable* state,
never on the in-memory roster:

- `Inserted` is returned only when `persist_named_groups_mutation` reports
  `AtomicWriteOutcome::Durable` — rename **plus** parent-directory fsync.
  `ReplacedNotDurable` is visible but unproven, and withholds.
- `Duplicate` calls `confirm_named_groups_durability` first and withholds if it
  fails.

**The campaign draft does not honor this, and the port fixes it.** Campaign
`admit_public_group_bootstrap` returns `Ok(Duplicate)` straight off a
`state.named_groups.read()` in two places (the already-installed branch and the
`NotReplaced` branch). A roster write that renamed into place but failed its
parent-directory fsync stays visible in memory with
`named_groups_requires_durability_confirmation` raised — so the campaign would
answer `Duplicate` there, and the sender would trade the last durable record of
the obligation for evidence that might not survive a power loss.

Guarded by `duplicate_completion_requires_confirmed_durability`, which asserts
the confirmation flag is **cleared** after a `Duplicate` — only possible if
admission actually forced the confirmation before answering. Verified by hand:
removing the `confirm_named_groups_durability` call makes it fail on
"Duplicate must not be answered until durability has been re-confirmed".

### Fail-closed boot contract (reviewer requirement 2)

**What happens.** `load_public_group_bootstrap_outbox` returns `Err` and
`serve_with_options` shuts down the exec service and the agent and returns a
startup error. The daemon does not come up: no API port file is published and no
listener task is registered, because the loader runs before both. Rejections:
the file is not valid JSON; `version != 1`; more than 1024 entries; a duplicate
key; any per-entry validation failure (payload over the DM size limit, payload
digest ≠ stored digest, key ≠ recomputed binding digest, decoded group's
`(stable id, revision, state_hash)` ≠ the obligation's, or the group is not
SignedPublic / is withdrawn / does not retain exactly its head commit); or a
first reconciliation that is not directory-durable. A **missing** file is not an
error — that is first boot with no outstanding debts.

**Why abort rather than limp on.** Every entry in that file corresponds to a
member who is *already on the durable roster* and is waiting for a snapshot they
cannot get any other way: a cold bootstrap is the only path into a group you do
not hold, since metadata commits only advance a group you already have. Dropping
an entry to keep booting does not degrade a nice-to-have — it strands that
member permanently.

It also strands them *invisibly*. Nothing reconstructs obligations from the
roster, so no later pass would notice the gap; the next reconciliation would
simply agree there is nothing owed. Limping on therefore converts a loud,
one-time, operator-visible startup failure into an unbounded membership bug that
surfaces days later as "why can't Alice see the group", with no evidence left
pointing at the sidecar.

The corruption cases are also not partially trustworthy. Digest mismatch, key
mismatch, and payload/frontier disagreement all mean the file's own bindings
contradict each other — precisely the state where you cannot tell which of two
disagreeing values is true. Proceeding would send one frontier's payload while
waiting for an ACK describing another, producing an obligation that can never be
cleared: a permanent 60-second retry loop against a live peer.

The cost of aborting is bounded and recoverable, because the sidecar is derived
state rather than source of truth. An operator who wants to proceed can delete
`public_group_bootstrap_outbox.json` and restart; the price is that affected
members must be re-added to regenerate their obligations. That is deliberate,
logged, operator-chosen loss of a durability promise — the only acceptable way
to lose one.

This is not a remote-DoS surface: the file is written only by this daemon, under
a persistence lock, via atomic rename. A peer cannot cause a malformed sidecar.
And it is the established convention here, not a new severity — ADR 0028's
causal-approval-queue and predecessor-relay-outbox loaders already abort startup
on a rejected sidecar, under the same "every fallible loader runs before the API
port is published" rule.

## Send entry point

Per the handover's argued decision, the campaign's
`send_direct_with_config_and_thread(..., Option<DmThreadMeta>, Option<DmLogicalId>)`
is **not** ported — it drags deferred v3 threading in. Instead:

```rust
// src/dm.rs — DmSendConfig
pub logical_request_id: Option<[u8; 16]>,
```
```rust
// src/dm_send.rs — send_via_gossip
let request_id = config.logical_request_id.unwrap_or_else(fresh_request_id);
```

The outbox sets it to the **first 16 bytes of the obligation's binding digest** —
the same bytes the obligation is keyed by, not a separate derivation. Obligation
and wire request are one identity, which is what makes a post-restart retry the
same logical request and lets an ACK be matched back to the obligation it
discharges. `DmDurableBindingDigest` / `complete_durable` are receiver-side
replay-cache constructs and cannot set a sender's wire request id, so they are
not used here.

## Line-count accounting

| File | Before | After | Δ |
|---|---|---|---|
| `src/server/routes/named_groups.rs` | 28,274 | 28,170 | **−104** |

Comfortably under the 28,274 bar ("prefer shrink"). Three things moved out: the
old fire-and-forget `spawn_public_group_bootstrap_delivery` (−47); the two new
call sites delegate to outbox-module helpers
(`public_group_bootstrap_obligation_for_add`,
`cancel_public_group_bootstrap_obligations_for_removal`) rather than inlining
outbox policy in the roster handlers; and `admit_public_group_bootstrap` (−98)
moved to the outbox module, which is where the completion-ordering fix belongs —
next to the handler that resolves the completion, and next to the outbox whose
obligation the ACK deletes. Visibility bumps
(`persist_named_group_info`, `write_named_groups_json_atomic`,
`named_group_direct_delivery_config`, `signed_public_bootstrap_snapshot`,
`validate_public_group_bootstrap`, `persist_named_groups_mutation`,
`confirm_named_groups_durability`, `MAX_BOOTSTRAP_INSTALLED_GROUPS`,
`PublicGroupBootstrap` fields) cost no lines.

## Omitted hunks, with evidence

1. **`send_direct_with_config_and_thread`** — replaced by
   `DmSendConfig::logical_request_id` (above).
2. **Raw-probe-then-gossip-repair split** (`public_group_bootstrap_raw_delivery_config`,
   `public_group_bootstrap_gossip_repair_config`, `deliver_*_with_configs`,
   `deliver_*_gossip_repair`). On main the raw path is guarded by
   `prefer_raw_quic_if_connected && !require_gossip && !require_durable_app_ack`
   (`src/lib.rs`, the `preferred_raw_receipt` binding), so a strict bootstrap
   never takes the raw probe and the campaign's two-config structure degenerates
   to two identical gossip attempts. Ported as one gossip delivery config.
3. **The successor-suffix admission family** (`public_group_bootstrap_successor_snapshot`,
   `_successor_chain_valid`, `_successor_state`, `_snapshot_matches`,
   `_retained_log_valid`, and the multi-commit relaxation of
   `validate_public_group_bootstrap`). This advances an *already installed*
   group from a verifiable commit suffix — a receiver-side admission feature,
   not the outbox, and the largest single block in the campaign commit.
   Reconciliation instead re-targets a stale obligation at the group's current
   frontier using main's existing `signed_public_bootstrap_snapshot`, which is
   what stops an obligation being retried forever at a frontier no ACK can
   match. Slice-4 candidate.
4. **`PublicGroupBootstrapDeliveryHookControl` / `take_public_group_bootstrap_delivery_hook`** —
   test scaffolding for the in-flight-fencing tests, which belong to (3)'s family.
5. **Hedged ACK publisher and raw durable frame transport** — slice 4, per the
   handover's evidence.
6. **ADR §6 documentation, partially.** `CHANGELOG.md` is updated in this PR,
   matching the convention every prior DA slice followed, and its entry names
   the new sidecar file. The rest of §6 (`docs/api-reference.md` for
   `/direct/send`'s `logical_id`, `docs/design/dm-over-gossip.md`, the ADR
   status link) is bound to slice 4's user-facing surface and is deferred with
   it. Note there is no existing sidecar-inventory doc to extend: the ADR 0028
   sidecar (`predecessor_relay_outbox.json`) is not documented under `docs/`
   either.

## Deviations from the handover / campaign

- **Durable route registration.** The campaign registers the bootstrap prefix
  with plain `with_typed_payload_route` (it predates 3a). That would withhold
  every v2 ACK by policy and livelock the outbox, so this PR uses
  `with_durable_typed_payload_route`. Verified by hand: downgrading it makes
  `serve_installs_signed_public_group_from_direct_bootstrap` fail — and it fails
  at *install*, not at the ACK, because ADR 0030 §7 refuses a non-opted-in typed
  route before dispatch, so the handler never sees the payload. The revert-guard
  comment records that observed behaviour.
- **Already-installed-at-a-different-frontier now withholds the ACK.**
  `admit_public_group_bootstrap` answers `Duplicate` only when the installed
  frontier matches the incoming one exactly; a receiver that still trails the
  authority returns `Err`, which withholds the ACK and keeps the sender's
  obligation alive until a retry sees a match. This is the literal reading of
  "cleared only by a frontier-matching v2 ACK", and it is the fail-safe
  direction: a visible pending obligation rather than a false clear. (On `main`
  this path returned silently and delivered nothing, so it is not a regression.)
- **A SignedPublic add now fails loudly if its snapshot cannot be built.**
  `public_group_bootstrap_obligation_for_add` returns `Err` and the handler
  answers 500, before any roster mutation is persisted (this matches the
  campaign). Previously the add succeeded and the bootstrap was silently
  skipped — which is the bug this slice exists to close: a member on the roster
  that nobody remembers to bootstrap.
- **`named_groups::tests` is now `pub(in crate::server)`** so the outbox unit
  tests reuse the existing `AppState` fixture instead of duplicating a
  ~150-field literal.
- **Test harness:** `start_daemon` retries `serve_with_options` for up to 30 s.
  A restart on the same data directory briefly races the previous instance
  releasing its SQLite history database (`database is locked`); this is a
  harness-only concern and the first start is still instant.

## Tests

`tests/f2_public_group_bootstrap_wiring.rs` is **updated, not replaced** (ADR
0025 / F2), and now drives the outbox path:

- `serve_installs_signed_public_group_from_direct_bootstrap` — existing test,
  now also asserts the authority's sidecar **drains**, so the ACK path is
  covered end to end. Its revert guard is rewritten to name all four
  `serve_with_options` wiring pieces this slice adds.
- `bootstrap_outbox_survives_sender_restart_and_clears_only_on_ack` — the ADR
  §5 validation, dispatch-level: the recipient is taken **down before** the
  member add, so no ACK can exist; the authority writes a durable obligation;
  several retry passes fail without clearing it; the **authority restarts** and
  the obligation is still on disk; the recipient returns and only then does the
  group install and the obligation drain.

Eleven unit tests in the new module cover completion ordering
(`duplicate_completion_requires_confirmed_durability`), key binding, request-id stability and
identity with the obligation key, backoff (including the shift-overflow edge),
restart survival through the sidecar, all four fail-closed load rejections, the
1024 cap, `LegacyV1Sent` rescheduling, failed-send rescheduling, roster-driven
reconciliation, and exact cancellation.

## Gates

All run with real exit codes.

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | exit 0 |
| `cargo clippy --workspace --all-features --all-targets -- -D warnings` | exit 0 |
| `cargo nextest run --workspace --all-features` | exit 0 — **2646 passed, 0 failed**, 260 skipped, in 95 s |
| `scripts/check-panics.sh` | exit 0 — zero unwrap/expect/panic in production code |

No flakes fired: neither documented quic-localhost flake
(`direct_send_with_require_ack_round_trips_to_live_peer`,
`suppressed_peer_inbound_redial_is_rejected`) needed a retry, so no isolated
re-run was required.

Do not merge — cross-review follows.


dirvine:
## Cross-review (slice 3b / ADR 0030 §5)

Walked `85b8079` against the 3b bar.

**Size.** `named_groups.rs` 28274 → **28268** (−6). Outbox is its own module. Bar met.

**Obligation.** Binding digest is the key and the first 16 bytes are `DmSendConfig::logical_request_id`. Same identity on disk and on the wire. `require_durable_app_ack = true`, `max_retries = 0` (outbox owns schedule). Cap 1024. Backoff `1s<<min(n,6)` clamp 60s.

**Completion.** Only `V2ApplicationAck` plus a still-current frontier clears. `LegacyV1Sent` and send errors reschedule. Trailing receiver (`admit` frontier mismatch) returns `Err` → ACK withheld → obligation stays. Missing capability does **not** downgrade to v1.

**Opt-in.** `with_durable_typed_payload_route`. Handler resolves `Inserted | Duplicate` only after consent + signed-frontier check + directory-durable persist. Campaign's plain typed route would withhold pre-dispatch — using the durable variant is required, not a taste call.

**Fail-closed load.** Version / over-cap / duplicate / payload-frontier mismatch abort `serve_with_options`. Argued in the PR body. Accept.

**No advert change.** Ceiling 2. No product-default flip.

**Deviations I accept:**
1. Durable route (required).
2. Successor-suffix omitted; reconcile re-targets to the current frontier. Slice-4 candidate is the right parking place.
3. One gossip config (raw is dead under `require_durable_app_ack`).
4. Duplicate only on exact frontier (fail-safe).
5. Add fails 500 if snapshot cannot be built, before roster mutation.
6. F2 updated, not deleted, plus restart-survive test.

**§6.** CHANGELOG names the sidecar and the fail-closed rule. Remaining api-reference / dm-over-gossip surface rides slice 4 as you said. Fine.

**Approve. Merge on CI green.** Same-account, no formal approve. Slice 4 still owns product defaults, `idempotency_conflict`, successor-suffix if you want it, and the rest of §6.

dirvine:
## Re-review of `438867f`

Previous glance was on `85b8079` only. This delta is a real gap, not a drive-by.

**The unsoundness is real.** `Duplicate` off the in-memory roster while `named_groups_requires_durability_confirmation` is raised would let the sender delete the last durable record of the debt against a write that may not survive power loss. Gating `Duplicate` on `confirm_named_groups_durability` — and withholding if confirmation fails — is the correct fail-safe. `Inserted` already required `AtomicWriteOutcome::Durable`.

**Test.** `duplicate_completion_requires_confirmed_durability` forces the confirmation flag, then asserts `Duplicate` *and* that the flag is cleared. That is the right observation: the flag can only be false if admission actually ran confirmation. Good.

**Move.** `admit_public_group_bootstrap` living in the outbox module is the right layer (it is the completion the obligation hangs on). `named_groups.rs` now 28,170.

**Fail-closed boot** write-up in the PR body is what I asked for.

**Approve `438867f`.** Merge when CI is green **on this SHA** (the 85b8079 Integration(groups) timeout re-ran green on the same commit — treat as flake unless the 438867f groups job fails the same way). Same-account, no formal approve. No merge-watcher until that job is green.

Issue #333, opened 2026-08-15
flaky: named_group_integration two-daemon suite fails intermittently under CI load (multiple tests, multiple assertion points)


dirvine:
Three Integration (groups) failures on 2026-08-15, all in tests/named_group_integration.rs, all at DIFFERENT points:

1. Run 31883474982 on **main** (dc3e862, slice-2 merge, pre-outbox): `named_group_creator_removal_propagates_to_removed_peer` panicked at :1339. Next main run green.
2. Run 31891071142 on PR #332 (85b8079): `named_group_admin_delete_propagates_to_peer_after_creator_delete_409` panicked at :1769 ("alice never observed bob's invite join", 30s wait_until). **Re-run of the identical SHA: green.**
3. Run 31891936441 on PR #332 (438867f): same test as (2) but panicked at :1899 ("recipient TreeKEM snapshot should be wiped after GroupDeleted") — got much further before dying.

Pattern: two-daemon convergence tests with 30s wait_until windows failing at varying points under CI load, on main and PRs, before and after the ADR 0030 outbox change — consistent with environmental timing flakiness, not a code regression (a deterministic regression fails at a stable point).

Same family as the quic-localhost findings (#316): multi-daemon loopback suites starved by ambient scheduler load. Candidate remedies, in order of evidence: nextest threads-required reservation for this binary (the gossip_plane_isolation precedent, PR #331), longer/adaptive wait_until windows, or serializing the suite.

Refs: ADR 0030 landing cycle; PR #331 (scheduler reservation precedent); PR #332 (where runs 2 and 3 fired).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

dirvine:
ACK from the DA cross-review: the three-point/two-test/one-on-main pattern is a flaky suite, not an outbox regression. Same family as #316. #332 can merge when 438867f Integration (groups) is green; the threads-required reservation is the #333 follow-up, not a 3b blocker.

dirvine:
Additional diagnosis from the slice-3b builder (full ignored-suite run on the #332 branch, 27 tests, --no-fail-fast, 386.8s):

- 25/27 passed — including the 24 tests CI's fail-fast cancel never reached, so no regression was hiding behind the cancel.
[...]

PR #334, opened 2026-08-15
feat(dm): DA slice 4 — product durable-by-default + logical_id + idempotency_conflict (ADR 0030 §4/§6)

dirvine:
> ### Status at head `8a23f47` — read this before reviewing an older SHA
>
> This branch moved several times during review. Verify against the head SHA;
> `c54d22e`, `3ff374a`, and `b291448` are all stale.
>
> | Pre-merge item | State at `8a23f47` |
> |---|---|
> | 1. Restore main's slice 1–3 Unreleased bullets | **Done.** main's entire Unreleased `Added` block (5175 chars, all 9 bullets) present verbatim, slice-4 bullets on top. Nothing condensed. |
> | 2. Two stale Notes | **Done.** Zero occurrences on head. |
> | 3. Calibrate the two durable-send test clients 10s → 30s | **Done**, per ruling (b'). No `~20s` sentence added to api-reference, per the same ruling. |
>
> **One residual, and it is a product finding rather than a test issue —
> see "Gate results".** Production code is unchanged since the approved
> `3ff374a` (`git diff --stat 3ff374a HEAD -- src/` is empty), so this stays
> in glance scope.

Completes the ADR 0030 durable-application-ACK campaign (slices 1–4). Precedes
the 0.38.0 release; **no version or tag changes here**.

## Scope

**1. Product send tier defaults (ADR §4).** `POST /direct/send` and
`x0x direct send` default `require_durable_app_ack = true`, with an explicit
opt-out (`require_durable_app_ack: false` / `--no-durable-ack`). The flip lives
in the request-scoped builder `direct_send_config_for_request`, **not** in the
shared `direct_message_send_config()` helper — the WS frame and every
control-plane send read that helper, so flipping it there would have made the
whole internal tier strict and reintroduced v0.37.0's welcome-blob livelock.
A test (`the_shared_internal_send_config_stays_v1`) pins that boundary.

**2. `require_gossip_ack` → hard 400.** Setting the field in any form answers
`400 require_gossip_ack_removed`. It is typed as `Option<serde_json::Value>`
rather than `Option<bool>` on purpose: `"require_gossip_ack": "maybe"` must get
the same documented 400 as `false`, not an opaque JSON-extractor rejection. An
explicit `null` is treated as omission, like every other optional field on this
body. `DmSendConfig::require_gossip_ack` is untouched for library callers.

**3. REST `logical_id`.** Validation ported from the campaign branch's
`DmLogicalId` (1–128 bytes of `[a-z0-9._:-]`, rejected rather than normalized).
Rejection happens before any send: silently dropping an unusable token would
hand the caller a fresh random id and the retry identity they asked for would
not exist.

**4. `idempotency_conflict`.** New `DmError::IdempotencyConflict` /
`DmAckOutcome::IdempotencyConflict` (postcard index 3, appended last — a
discriminant test pins 0/1/2/3). Both receiver binding-conflict sites — the
in-memory replay cache and the restart-spanning durable-history lookup — now
report it instead of reusing `AckSemanticsUnavailable`, which is the reuse the
#329 review flagged as "least-wrong until this error lands".
`AckSemanticsUnavailable` keeps only its semantics-unavailable meaning.

## Review round (head is `3ff374a`)

Three requested changes, all in:

1. **GUI migration** (below) — the ship-blocker.
2. **404/409 boundary widened.** Was `contacts.get(to).is_some_and(|c|
   c.dm_capabilities.is_some())`, which still filed a pre-capability card as a
   stranger. Now `contacts.get(to).is_some()`: any known card ⇒ 409, only a
   wholly unknown id ⇒ 404. Pinned by
   `a_known_contact_without_capabilities_is_a_409_not_a_404`.
3. **`logical_id` + durable opt-out ⇒ 400 `logical_id_requires_durable_ack`.**
   Only the durable receiver path consults the id, so the pair would hand back
   an at-least-once retry identity nothing enforces — discoverable only once
   duplicates appear. `x0x direct send` rejects `--logical-id` with
   `--no-durable-ack` via clap, before the round trip. The docs were also
   re-ordered per review: the restart-spanning promise was stated four lines
   *before* the durable precondition that makes it true, so the constraint now
   opens the field definition and the reason is given — every guarantee in that
   section is made by the recipient's durable path.

Also reverted the hedged ACK publisher back out (see deferred families) — this
PR's `src/dm_inbox.rs` and `src/direct.rs` are byte-identical to their state
before that port attempt.

## Migration: the hard-400's blast radius

The 400 is a ship-blocker for any caller that sets the field, so I grepped the
whole tree for producers before enforcing — and found one:

`src/gui/x0x-gui.html` set `body.require_gossip_ack=true` on `/direct/send`
whenever the DM "ACK" checkbox was ticked. **Note the shape**: a quoted-key
search (`"require_gossip_ack"`) does **not** find it, because the GUI assigns
via dot notation. `git grep` for the bare identifier does. Migrated by removing
the field — nothing replaces it, because the product default is now durable
delivery, which is strictly stronger than the gossip ACK the checkbox was
requesting. The checkbox keeps its remaining job, the post-send liveness probe.

Every other hit in the tree is a Rust `DmSendConfig` field assignment
(`stores.rs`, `named_groups.rs`, `exec/service.rs`) — library-internal config,
not a REST body, and explicitly unaffected: `DmSendConfig::require_gossip_ack`
is unchanged.

While in the GUI I added a `/direct/send` error map, because ADR 0030 calls the
409 UX "load-bearing, not transitional" and the GUI is a product surface. It
renders `recipient_ack_semantics_unavailable` as "Peer needs upgrading… retry
later, or resend without durable delivery" and `idempotency_conflict` as
"Retrying won't help — send it as a new message", instead of showing the raw
code for a condition where retrying is the one thing that cannot work.

`gui_smoke` (11/11) passes.

## Interop sanity check

Durable-by-default is only safe if two stock daemons interoperate.
`[history].enabled` is **default-on** in the daemon (`server/state.rs`), and the
advert is v2-iff-history, so two stock 0.38 daemons negotiate durable ACKs. The
409 is reached by genuinely not-yet-upgraded peers and by daemons that
explicitly set `history.enabled = false` — documented in api-reference.

## Defaults matrix

| Surface | Tier | `require_durable_app_ack` | `logical_id` |
|---|---|---|---|
| `POST /direct/send` | Product | **`true`** (opt out with the field) | accepted |
| `x0x direct send` | Product | **`true`** (opt out with `--no-durable-ack`) | `--logical-id` |
| WS `send_direct` frame | Internal | `false` | not accepted |
| `Agent::send_direct*` default config | Internal | `false` | via `DmSendConfig` |
| Daemon control plane (welcome/TreeKEM/metadata) | Internal | `false` | per named config |
| Bootstrap outbox (§5) | Internal, explicitly strict | `true` | obligation digest |

The CLI sends `require_durable_app_ack` **explicitly** rather than relying on
the daemon default, so a version-skewed CLI/daemon pair can never silently swap
receipt semantics.

## logical_id derivation

```
request_id = blake3::derive_key(
    "x0x dm logical request id v1",
    sender_agent_id ‖ recipient_agent_id ‖ logical_id,
)[..16]
```

Ported from the campaign's `dm_send::logical_request_id` rather than the
simpler blake3-of-token the brief allowed. Both agent ids are bound. The
recipient dedupes on `(sender, request_id)`, so binding the sender is
belt-and-braces; binding the **recipient** is load-bearing — the sender's
in-flight ACK waiter is keyed by `request_id` alone, so one token fanned out to
two peers must not produce one id that two waiters contend for. The token never
goes on the wire.

## Two fixes the default flip forced

Both were found by the gates, not by inspection, and neither is cosmetic.

**The conflict check was unreachable through real ingress.** `handle_incoming`
re-ACKed any envelope already in the replay cache *before* comparing its
durable binding — so a v2 caller reusing a `logical_id` for different bytes was
told `Accepted` for content the receiver never stored, and the §1 binding check
at `dm_inbox.rs` was dead code on that path. Durable envelopes replaying a
completed request now fall through to the durable path, which owns the whole
replay decision. Narrowly scoped via `cached_completion_needs_binding_check`:
only a v2 payload replaying a completion that *has* a durable binding pays the
signature-verify + decrypt; v1 replays and ACK envelopes keep the cheap path.
This was latent before slice 4 — only the bootstrap outbox set a logical
request id, and it derives one per payload digest, so a rebind was impossible
by construction. Exposing `logical_id` to callers is what makes it reachable.

**A stranger was reported as a peer needing an upgrade.** The strict gate fired
before any key-availability check, so with every product send now strict, a
send to an undiscovered agent id would have returned `409
recipient_ack_semantics_unavailable` instead of `404`. That tells a product UI
to chase a fleet-upgrade problem it does not have. `AckSemanticsUnavailable`
now requires that we actually know the peer — an advert **or any contact card**
(`cap` alone cannot answer this, because the strict path deliberately withholds
the card fallback). Per review, the boundary is "do we know this agent", not
"has its advert converged": a card with `dm_capabilities: None` is a known peer
mid-convergence, plausibly one needing an upgrade, so it gets the 409. Caught by
`daemon_api_direct_send_not_found`, an Integration-only ignored test that a
`--workspace` run never executes.

## Two port items evaluated and declined, with evidence

**`HistoryCommitFailed` → `503 history_commit_failed`: not ported.** On the
campaign branch every raise site (`lib.rs:5108-5136`) is inside a **sender-side
durable-loopback history write** that main does not implement, and that code
also carries `thread_root`/`thread_parent` — v3 fields explicitly out of scope.
Main's loopback path (`send_direct_with_config_inner`) writes no history and
cannot fail this way, and a *receiver's* commit failure withholds the ACK by
design (§1), so the sender sees a 504 timeout, never a typed error. Adding the
variant would create an error nothing can construct and document a REST code
the daemon can never emit. Flagging rather than silently skipping.

**Product timeouts: no change, and the reason is not laziness.**
`PRODUCT_DM_ATTEMPT_TIMEOUT = 8s` already matches main's
`direct_message_send_config()`. `PRODUCT_DM_TOTAL_TIMEOUT = 30s` has no
equivalent field in main's `DmSendConfig`, and it is not needed: `max_retries:
1` with `ExponentialFromTimeout { factor: 2 }` gives 8 s + 8 s backoff + 8 s ≈
24 s worst case, already inside 30 s. Adding a total-timeout field would be new
machinery ADR §4/§6 does not ask for. Raising
`raw_quic_receive_ack_timeout` from 8 s to 30 s would be an active regression:
raw QUIC is skipped entirely for durable sends, so it only affects opted-out
sends — where it would make a dead raw path hang 30 s instead of failing at 8 s
and falling back — and it lives in the **shared** helper, so it would hit WS and
every control-plane send too.

## Deferred families — decisions

**(a) Hedged multi-route ACK publisher — DEFERRED to #335.** This briefly
landed and has been reverted out of this PR at the reviewer's direction, which
I agree with: mixing a change to *when the ACK reaches the wire* with a
product-default flip is exactly what a release rider should not do, and the
saturated-queue drop interacts with the commit-before-ACK promise in a way that
deserves its own review and validation-matrix run.

The evidence for deferring, for the record: cherry-picking each commit onto
`origin/main` in an isolated worktree conflicts every time (`21387d2` in
`dm_inbox.rs`; `bde4d28` in `dm_inbox.rs`, `direct.rs`, and the design doc;
`fa538eb` in `dm_inbox.rs`) — ~+670/−160 restructuring `DmInboxService`,
`InboxPipeline`, and `spawn_subscription_loop`, the code slices 2/3/4 rewrote.

The gap it addresses is real and tracked in **#335**: main awaits
`publish_ack_for_protocol` inline at 8 call sites in the inbox loop with no
timeout, and durable-by-default increases v2 ACK traffic. `src/dm_inbox.rs` and
`src/direct.rs` in this PR are byte-identical to their `c219e4b` state.

The ported implementation and its three behavioural tests are preserved on
`wip/335-hedged-ack-publisher`, with a take-this/skip-that porting note on #335
so the work becomes a ready-made PR there rather than being redone.

**(b) Raw durable DM frame transport — DEFER.** `handle_raw_frame` is
inseparable from the `AckRoute` / `handle_incoming_routed` refactor — one hunk
of `6a58d81` (+1866/−174), the same refactor the survey says to avoid taking.
Independently, `RAW_DURABLE_DM_MAGIC` is a **new wire framing**, and ADR 0030
states the entire compatibility surface is "ACK semantics + capability adverts"
with envelope layout unchanged. Post-0.38.

**(c) Successor-suffix admission — DEFER.** The two validators are liftable
pure functions (~110 lines), but every call site is entangled with the
`4880a9f` full-file rewrite, and the 26 call-site references live in
`named_groups.rs` — the 33.7k-line file whose size ADR 0030 §5 explicitly
constrains ("must not grow past its pre-branch size"). Hand-porting the two
validators plus their call sites is the path for post-0.38; doing it here would
grow exactly the file the ADR forbids growing.

## Docs (ADR §6)

- `docs/api-reference.md` — request body, `logical_id`, the durable default and
  its rollout consequences, the hard 400, all new error codes, the 404-vs-409
  precedence, CLI flags, WS-is-internal note, and the bootstrap outbox sidecar
  (`<data_dir>/public_group_bootstrap_outbox.json`, fail-closed startup).
- `docs/design/dm-over-gossip.md` — send-surface tier table, `logical_id`
  derivation, corrected receiver-ordering step 2/3 with the two distinct
  outcomes, and the fast-path stand-aside rule.
- `CHANGELOG.md` — Unreleased, 0.38.0 framing, both breaking changes called out
  as BREAKING with the migration, plus the hedged publisher and the new
  diagnostics counter.
- **Mixed-version matrix rows this slice owns** (`dm-over-gossip.md`): a
  receipt-semantics matrix (0.38 strict / 0.38 opted-out / internal tier / 0.37
  sender × v1-advertising vs v2-advertising recipient) and a precedence table
  covering the unknown-recipient 404, same-id-same-bytes vs same-id-different-bytes,
  the `require_gossip_ack` 400, and the withheld ACK on commit failure.
- The bootstrap outbox sidecar is now **named** in `dm-over-gossip.md`
  (`<data_dir>/public_group_bootstrap_outbox.json`), not only in api-reference.

**ADR §6 item with nothing to do:** the daemon config keys
`dm_capability_cache_ttl_secs` and `dm_capability_test_controls` do not exist
in the tree (`grep` over `src/` finds zero hits) — slice 1 did not port them,
so there is no key to document. Flagging rather than silently skipping.

## Gate results

All run locally with real exit codes on `8a23f47`.

| Gate | Result |
|---|---|
| `cargo fmt --all -- --check` | exit 0 |
| `cargo clippy --workspace --all-features --all-targets -- -D warnings` | exit 0 |
| `cargo nextest run --workspace --all-features` | **2655 passed**, 0 failed, 260 skipped |
| `scripts/check-panics.sh` | exit 0 — zero panic/todo/unimplemented in production |

**Ignored / CI-Integration-only suites** (the slice-3b lesson — a `--workspace`
run does not execute these):

| Suite | Result |
|---|---|
| `--test api_manifest` | 3/3 passed |
| `--test api_coverage --run-ignored all` | 16/16 passed |
| `--test daemon_api_integration --run-ignored ignored-only` | **100/100 passed** — includes `daemon_api_direct_send_not_found`, which the precedence fix above was required for |
| `--test history_wiring --run-ignored all` | 4/4 passed |
| `--test peer_lifecycle_integration --run-ignored all` | 6/6 (see the finding below — one failure in seven runs total) |
| `--test x0x_0041_prefer_newest_test --run-ignored all` | 4/4 passed |
| `--test x0x_0041_synthetic_kill_restart --run-ignored all` | 2/2 passed |
| `--test gui_smoke --run-ignored all` | 11/11 passed — covers the migrated GUI |
| `--test named_group_integration --run-ignored ignored-only` | 25/27 passed, **2 failed — pre-existing** |
| `--test history_api --run-ignored all` | 1/3 passed, **2 failed — pre-existing** |

**Both failing suites verified against a clean `origin/main` (f16448f)
checkout, not assumed:**

- `named_group_integration`: `named_group_admin_delete_propagates_to_peer_after_creator_delete_409`
  and `named_group_creator_removal_propagates_to_removed_peer` fail identically
  on `origin/main` with the same message ("bob never applied alice's
  authoritative member add before removal"). Known flaky family, issue #333.
  Both propagate over `named_group_direct_delivery_config()` — internal tier,
  untouched by this diff.
- `history_api`: `history_backpressure_smoke` fails on `origin/main` and on
  every run here. The *second* failure moves between runs — `ws_backfill_then_live_no_gap_no_dup`
  on the main baseline and on the latest run here, `rest_history_list_search_stats_purge_roundtrip`
  on an earlier run — which is itself the flakiness signature. This suite drives
  **topic publish** ingest, not DM sends; nothing in this diff touches that path.

### One residual — a product finding, not a test flake

`peer_lifecycle_integration` fails roughly **1 run in 6**. Raising the client
timeout to 30 s (ruling b') removed the *client-side* timeouts, which finally
let the daemon's own answer be observed. It is not a 409, and not a harness
artifact:

```
direct/send 504 Gateway Timeout:
{"detail":"timed out after 1 retries over 24.002880959s","error":"timeout","ok":false}
```

- **Not** `409 recipient_ack_semantics_unavailable` ⇒ the advert had converged
  and the strict gate passed. This is not the convergence race we assumed, so
  no `wait_until` can close it.
- **504 after 24.0028 s** ⇒ the daemon exhausted its whole budget (8 s + 8 s
  backoff + 8 s) waiting for the **v2 application ACK**, against a live peer on
  loopback.

The receiver plausibly committed and dispatched and the ACK never got back —
the "committed but never acked" outcome, which is precisely what the hedged
multi-route ACK publisher deferred to **#335** exists to remove. Filed with
full detail on **#336**, including the note that #335 may be more load-bearing
than "ACK liveness, nice to have": durable-by-default is what makes this
reachable, and it reproduces on loopback with two healthy daemons.

Both status assertions now print status + body, so the next occurrence is
self-diagnosing. I have deliberately not masked this with a retry loop or a
further timeout widen.

No API-manifest regeneration was needed: routes, `cli_name`s, and descriptions
are unchanged, and the manifest carries no request-field schema. `route_set_matches_registry`
and `manifest_matches_registry` both pass.

## Omitted hunks

Nothing v3 — no `DmThreadMeta`, no `DM_PROTOCOL_THREADED`. The receive ceiling
stays `DM_PROTOCOL_DURABLE_ACK` (2). All campaign material was taken
hunk-by-hunk from individual commits; no file blobs, no `checkout --theirs`.

## Tooling note

`grep` in this shell is a ugrep wrapper (verified: `ugrep 7.5.0`, skips binary
files, different glob handling from `/usr/bin/grep`). All campaign-branch
surveys used real `git grep` / `git show` against explicit revisions, and the
portability claims above come from actual `git cherry-pick` runs in a throwaway
worktree, not from reading diffs.

Do not merge — review first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,21 @@ All notable changes to this project will be documented in this file.
 - **`idempotency_conflict`** (`DmError::IdempotencyConflict`,
   `DmAckOutcome::IdempotencyConflict`, REST **409 `idempotency_conflict`**).
   Reusing a `logical_id` for different bytes is now its own typed error.
+- **Hedged, bounded durable-ACK publication.** A v2 ACK is now published from a
+  background worker (queue 256, ≤32 in flight) that polls the recipient's
+  targeted inbox topic and the compatibility bus concurrently, each under a
+  22 s deadline, instead of awaiting both inline on the inbox loop. A targeted
+  publish can deliver remotely yet stay pending under per-topic backpressure,
+  which previously let a receiver commit and dispatch a DM while the sender
+  burned its entire ACK budget waiting on the reverse route. Commit-before-ACK
+  ordering is unchanged — only the publication is asynchronous. Durable-by-
+  default sends make this liveness path load-bearing, which is why it lands
+  with them. **v1 ACK publication is byte-for-byte unchanged**: still inline,
+  still targeted-only unless the payload arrived on the bus.
+- **`ack_publish_route_failed` on `GET /diagnostics/dm`.** The externally
+  visible signal that a committed message's ACK never reached its sender —
+  either a route failed or timed out, or the publisher queue was saturated.
+  Those senders see `504 timeout` and retry; nothing is silently lost.
 
 - **Durable signed-public bootstrap outbox (ADR 0030 §5, slice 3b).** Adding a
   member to a SignedPublic group used to direct-send the roster snapshot
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,35 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+> The ADR 0030 durable-application-ACK campaign completes in this release
+> (slices 1–4). The headline behaviour change for anyone building on x0x is in
+> **Changed** below: `POST /direct/send` is now durable-by-default, so a `200`
+> means the recipient committed the message, and a peer that has not upgraded
+> answers `409` instead of accepting silently.
+
 ### Added
+
+- **Product send tiers (ADR 0030 §4, slice 4).** The two send tiers named by
+  the ADR are now real. The *product* tier — `POST /direct/send` and
+  `x0x direct send` — requires ADR 0030 durable application-ACK semantics by
+  default. The *internal* tier — the WebSocket `send_direct` frame, the library
+  `Agent::send_direct*` default config, and daemon control-plane sends (welcome
+  blobs, TreeKEM plumbing, group metadata) — keeps v1 semantics; those messages
+  deliberately race connection establishment and livelock under strict gating.
+  `api-reference.md` marks the WebSocket surface internal.
+- **`logical_id` on `POST /direct/send` and `x0x direct send --logical-id`.**
+  A caller-supplied idempotency key (1–128 chars of `[a-z0-9._:-]`) giving
+  product callers the same at-least-once retry identity the bootstrap outbox
+  uses internally. Resending the same token to the same recipient — across a
+  timeout, a reconnect, or a restart — is one logical request: the recipient
+  re-ACKs the original commit rather than storing a duplicate. The wire request
+  id is `blake3::derive_key("x0x dm logical request id v1", sender ‖ recipient
+  ‖ token)[..16]`; the token itself never goes on the wire, and binding both
+  agent ids keeps one token fanned out to two peers from colliding on the
+  sender's ACK waiter.
+- **`idempotency_conflict`** (`DmError::IdempotencyConflict`,
+  `DmAckOutcome::IdempotencyConflict`, REST **409 `idempotency_conflict`**).
+  Reusing a `logical_id` for different bytes is now its own typed error.
 
 - **Durable signed-public bootstrap outbox (ADR 0030 §5, slice 3b).** Adding a
   member to a SignedPublic group used to direct-send the roster snapshot
@@ -83,6 +111,60 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- **BREAKING — `POST /direct/send` and `x0x direct send` are durable by
+  default (ADR 0030 §4, slice 4).** A `200` now means the recipient daemon
+  durably committed the message to its ADR-0023 history store and completed
+  local dispatch, not merely that it accepted the envelope. Pass
+  `require_durable_app_ack: false` (CLI: `--no-durable-ack`) for the previous
+  v1 semantics.
+
+  **What breaks:** a send to a peer running 0.37.x, or to any peer without
+  durable history enabled, now answers **409
+  `recipient_ack_semantics_unavailable`** where it previously returned `200`.
+  This is deliberate — ADR 0030 §2 forbids a silent downgrade, because a `200`
+  that sometimes means "committed" and sometimes means "enqueued" is the
+  ambiguity the whole protocol exists to remove. Handle the 409 as a
+  first-class UX state ("peer needs upgrade"), or opt out per message. The
+  refusal is fast and deterministic (one forced capability refresh), never a
+  timeout.
+
+  Durable sends also never take the raw-QUIC fast path, since a transport
+  receipt cannot certify a durable commit; `prefer_raw_quic_if_connected` is
+  ignored unless you also opt out of durable semantics.
+- **BREAKING — `require_gossip_ack` is rejected with 400 on `POST
+  /direct/send`** (ADR 0030 §4). Deprecated in the previous release, the field
+  now answers **400 `require_gossip_ack_removed`** when set in any form; an
+  explicit `null`, like omission, is still accepted. Accepting it as a silent
+  no-op was rejected in review: a client passing `require_gossip_ack: false`
+  for fire-and-forget would otherwise get a blocking durable send with no
+  signal that its request had been reinterpreted. Use
+  `require_durable_app_ack` to choose receipt semantics. Internal library
+  callers are unaffected — `DmSendConfig::require_gossip_ack` is unchanged.
+- **The strict-send gate no longer reports a stranger as a peer needing an
+  upgrade.** `AckSemanticsUnavailable` means "this peer does not advertise
+  v2", which presupposes we know the peer; an agent id with no capability
+  advert *and* no contact card now yields `RecipientKeyUnavailable` (404) as
+  it always did for non-strict sends. Without this, flipping the product
+  default would have turned every send to an undiscovered agent into a 409
+  telling the UI to chase a fleet upgrade for a peer that was never found.
+- **A rebound logical request now reports `idempotency_conflict`, not
+  `recipient_ack_semantics_unavailable`.** Both receiver-side binding-conflict
+  paths (the in-memory replay cache and the restart-spanning durable-history
+  lookup) previously reused the semantics-unavailable outcome, which told a
+  product "the peer needs upgrading" when the truth was "your client reused an
+  idempotency key". The two errors prescribe opposite repairs — retry-later
+  versus never-retry-these-bytes. `AckSemanticsUnavailable` now carries only
+  its semantics-unavailable meaning.
+- **A v2 replay is no longer re-ACKed before its binding is checked.** The
+  inbox fast path re-ACKed any envelope whose logical request was already in
+  the replay cache, which for a durable envelope meant answering `Accepted`
+  without comparing the payload against the committed binding — so a caller
+  that reused a `logical_id` for different bytes was told the new bytes were
+  stored when they were not. Durable envelopes replaying a completed request
+  now fall through to the durable path, which owns the full ADR 0030 §1 replay
+  decision. This was unreachable before slice 4 (only the bootstrap outbox set
+  a logical request id, and it derives one per payload digest); exposing
+  `logical_id` to callers is what makes it reachable.
 - **The v2 capability advert is now live and conditional (ADR 0030 §3).**
   `start_dm_inbox` advertises `max_protocol_version = 2` **iff** a durable
   history handle is present, else 1. The slice-1 hold is lifted because the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,96 +33,11 @@ All notable changes to this project will be documented in this file.
 - **`idempotency_conflict`** (`DmError::IdempotencyConflict`,
   `DmAckOutcome::IdempotencyConflict`, REST **409 `idempotency_conflict`**).
   Reusing a `logical_id` for different bytes is now its own typed error.
-- **Hedged, bounded durable-ACK publication.** A v2 ACK is now published from a
-  background worker (queue 256, ≤32 in flight) that polls the recipient's
-  targeted inbox topic and the compatibility bus concurrently, each under a
-  22 s deadline, instead of awaiting both inline on the inbox loop. A targeted
-  publish can deliver remotely yet stay pending under per-topic backpressure,
-  which previously let a receiver commit and dispatch a DM while the sender
-  burned its entire ACK budget waiting on the reverse route. Commit-before-ACK
-  ordering is unchanged — only the publication is asynchronous. Durable-by-
-  default sends make this liveness path load-bearing, which is why it lands
-  with them. **v1 ACK publication is byte-for-byte unchanged**: still inline,
-  still targeted-only unless the payload arrived on the bus.
-- **`ack_publish_route_failed` on `GET /diagnostics/dm`.** The externally
-  visible signal that a committed message's ACK never reached its sender —
-  either a route failed or timed out, or the publisher queue was saturated.
-  Those senders see `504 timeout` and retry; nothing is silently lost.
-
-- **Durable signed-public bootstrap outbox (ADR 0030 §5, slice 3b).** Adding a
-  member to a SignedPublic group used to direct-send the roster snapshot
-  fire-and-forget: an offline recipient never received it and nothing on the
-  authority remembered the debt. The send is now a durable obligation, in its
-  own module (`src/server/routes/public_group_bootstrap_outbox.rs`). It is
-  written in the same step as the roster commit, keyed by
-  `(recipient, group, frontier, payload-digest)` as one blake3 binding digest,
-  capped at 1024 entries, retried on `1s << min(attempts, 6)` clamped to 60 s,
-  and persisted to `public_group_bootstrap_outbox.json` in the data directory.
-  That sidecar is loaded **fail-closed**: an unsupported version, an over-cap
-  file, a duplicate key, or a payload that contradicts the frontier it claims
-  aborts daemon startup rather than silently dropping a promised delivery. An
-  obligation is discharged **only** by a v2 application ACK for that exact
-  frontier — released by the recipient's durable typed route after the consent
-  gate and a directory-durable install. A legacy v1 send and a failed send both
-  reschedule; neither is completion.
-- **`DmSendConfig::logical_request_id`.** An optional caller-supplied stable DM
-  request id; `None` (the default, and every existing caller) still draws a
-  fresh random one. The bootstrap outbox sets it to the head of its obligation's
-  binding digest, so obligation key and wire request id are one identity: a
-  retry after a sender restart is the *same* logical request, which is what lets
-  the recipient re-ACK instead of re-delivering and lets an ACK be matched back
-  to the obligation it discharges.
-- **Typed-route durable completion signal (ADR 0030 §7, slice 3).** Slice 2
-  withheld every v2 ACK on a typed route because typed-prefix families
-  classify `Ephemeral` and cannot be backed by a DM-history commit. A handler
-  can now report `DmTypedPayloadCompletion::{Inserted, Duplicate}` on a
-  oneshot carried in `DmTypedPayload`, and that signal releases the ACK.
-  `DmTypedPayload` also carries the envelope `request_id` so handlers can
-  dedupe on `(sender, request_id)` across restart, and is no longer `Clone`
-  (a oneshot sender has no meaningful copy). Routes opt in via
-  `DmInboxConfig::with_durable_typed_payload_route`; routes that do not opt in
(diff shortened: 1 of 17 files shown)


dirvine:
## Cross-review (slice 4 / ADR 0030 §4/§6)

Walked `c219e4b`. Most of the slice is the right shape. **Not approve yet.** Three things, one of them a ship-blocker I put on the bar last turn.

### Change requested

**1. Bundled GUI still sends `require_gossip_ack` (ship-blocker).**
`src/gui/x0x-gui.html:2852`:
```
if(ackBox&&ackBox.checked){body.require_ack_ms=5000;body.require_gossip_ack=true}
```
That is REST `POST /direct/send`. After this PR, ticking ACK 400s `require_gossip_ack_removed`. The PR body said no producers existed; this one does. The bar was: GUI/exec callers must be updated **in this same PR**. Exec is library `DmSendConfig` (fine). The GUI is a REST client. Drop `require_gossip_ack` from the body; keep `require_ack_ms` if the probe is still what the checkbox means.

**2. Known contact with `dm_capabilities: None` is classified as a stranger.**
`recipient_is_known = cap.is_some() || contact.dm_capabilities.is_some()`. A pre-capability contact card is a peer we know. Product send then 404s instead of 409 "peer needs upgrade / opt out" — the mixed-version case. Use `contacts.get(to).is_some()` (or trust-known), not "card embeds a capability blob". Add a test next to `a_strict_send_to_a_wholly_unknown_peer_…`.

**3. `logical_id` + `require_durable_app_ack: false` does not keep the promise.**
v1 replay does not bind id→bytes. Reuse with different bytes `Accepted`s while the cache is warm. Either **400** `logical_id` unless durable, or the docs must say the token is a no-op on the opt-out path. Prefer the 400 — a token you asked to be an idempotency key should not silently become a random-id cousin.

### Accept

- Flip lives in `direct_send_config_for_request`, not `direct_message_send_config()`. Test pins it. Welcome-blob livelock avoided.
- CLI sends the durable field explicitly. `--no-durable-ack` / `--logical-id` look right.
- `require_gossip_ack` 400 for any non-null JSON value; null ≡ omit. Good.
- `logical_id` derivation binds sender+recipient. Recipient binding is load-bearing for the waiter. Good.
- `IdempotencyConflict` postcard 3, both conflict sites rebound off `AckSemanticsUnavailable`. Discriminant test updated.
- Fast-path stand-aside so the binding check is reachable. The two-lookup race Greptile named is real but narrow; fold a single-lookup into the GUI/contact fix if it's cheap.
- Hedged ACK / raw frame / successor-suffix deferrals: agreed, with the evidence. File the 0.38.x ACK-publish follow-up.
- Ignored suites reported. `named_group_integration` / `history_api` fails match main.
- No 0.38.0 tag. `named_groups.rs` +5 (test fixtures). Fine.
- §6 keys that don't exist: correctly flagged.

Re-review when 1–3 are on the branch. Same-account, no formal request-changes.

dirvine:
## Re-review of `c54d22e` — ruling, not an approve

Walked the SHA independently of the builder report.

### Ruling: strip the publisher (path a)

`c54d22e` folds the #335 hedged ACK publisher into slice 4. I already ruled that out at 16:50Z. I am not going to re-open it on the merits inside this PR.

- Commit-before-ACK is about more than "history is awaited before `try_send`". A saturated queue drops a promised ACK after a durable commit. That is a contract change, and it is why #335 exists.
- v1-unchanged + 22s tests are relevant **on #335**, not here.
- Do not take `6a58d81`. That warning still stands when #335 is actually ported.

Revert `src/dm_inbox.rs` / the `ack_publish_route_failed` counter / the CHANGELOG hedged-ACK bullets / any publication-section docs that assume the worker. Leave https://github.com/saorsa-labs/x0x/issues/335 open.

### Item 1: done

GUI no longer sends `require_gossip_ack`. Checkbox keeps `require_ack_ms`. `DM_SEND_ERRORS` map is fine. Keep it.

### Items 2 and 3: not on this SHA

`c54d22e` does not touch `src/lib.rs` or `src/server/routes/direct.rs`. The "three bar items done" claim is incorrect.

**2. Known contact, `dm_capabilities: None` is still a stranger.**

```
recipient_is_known = cap.is_some() || contact.dm_capabilities.is_some()
```

A pre-capability contact card still 404s. The 404 row in `docs/api-reference.md` already says "no contact card" — the code does not match that sentence. Use `contacts.get(to).is_some()`. Add a test next to `a_strict_send_to_a_wholly_unknown_peer_reports_missing_key_not_missing_v2`.

**3. `logical_id` + `require_durable_app_ack: false` still derives a wire id.**

Not a 400, and the docs still describe the token as a restart-spanning idempotency key (a durable-path promise). That is neither option from the last review. 400 the combo. `idempotency_conflict` already being typed is a different item.

### Next

Re-push: 1 (already in) + 2 + 3, publisher out. Confirmed-final + ignored-suite report. I will re-review that SHA.

No merge. No 0.38.0 tag.

dirvine:
## Glance of `b291448` + cold-path ruling

Docs items check out. `src/` is empty vs `7db6f02`. Changelog slice 1–3 bullets are back; stale Notes are rewritten.

### Ruling: (b'), not (a)/(b)/(c)

**(c) no.** Do not merge with `peer_lifecycle` intermittently red.

**(b) as written, no.** Do not document "first durable send may take ~20s" in `api-reference` as the contract. That number is the gossip-inbox cold success path. Before this PR, product send preferred raw QUIC (`prefer_raw_quic_if_connected: true`) on a live connection — which this test fixture has — so the 10s HTTP client was calibrated to that path. Durable sends correctly skip raw QUIC (a transport receipt cannot certify a commit), so the first POST now pays publish + reverse application ACK. The daemon budget is already `timeout_per_attempt = 8s`, `max_retries = 1`, exponential backoff from that timeout (`≈ 8 + 8 + 8 = 24s`). tic-tac-toe already uses `DIRECT_SEND_REQUEST_TIMEOUT = 30s` for 0.38. Writing a 20s SLA would freeze a limitation as a feature.

**(a) not in this PR.** Your wait already returned before the POST timed out. `capability_store_entries > 0` is a count, as you said; advert warmup might skip the 3s forced refresh and does **not** obviously eat the remaining ~14s of gossip ACK. That needs a stage breakdown (gate vs publish vs ACK) and its own review. File it next to #335 with the 17,476 / 439 / 191 ms numbers. Do not rider it.

### (b') — then merge

1. Bump the two `peer_lifecycle` HTTP clients **10s → 30s**. That matches the daemon budget and the ttt client. It is not a `wait_until` widen. Keep the advert-wait; the comment already admits it does not close the gap.
2. File the 17.5s cold success path as a **0.38.x issue**, not an api-reference SLA.
3. Merge on CI green. No 0.38.0 tag from this PR.

Tag-time fact for David: ttt first-move-to-a-new-opponent will succeed (30s client) and feel ~17s until the follow-up.

I will not re-review unless `src/` changes.

dirvine:
Superseded by #339: the slice-4 content reviewed here at 8a23f47, rebased onto main with the #338 publisher (applied as this branch's net diff, per-file stat verified identical). This branch is left untouched as the review audit trail; #334 will be closed when #339 lands. The ruling-(A) validation result is in #339: the 1-in-6 durable 504 is gone (20/20) with the publisher underneath the flip.

